PR #21, opened 2022-02-02
Use normal methods for blocking I/O

d-j-hatton:
FastAPI should run blocking I/O (such as SQLAlchemy queries) faster if it is not in an asynchronous method. This is because it will run non-asynchronous path operation functions in a thread which it then `await`s. See the bottom of https://fastapi.tiangolo.com/async/ and the "About `def` vs `async def`" section of https://fastapi.tiangolo.com/tutorial/sql-databases/


Anthchirp (approved):
Good spot! Thanks

PR #28, opened 2022-02-09
Web interface

HorstmannA:
First pass at a web interface using Jinja2 templates.


Anthchirp (approved):
Would recommend moving templates out of the `server/ ` directory and putting them next to it.

Instead of using `__file__` magic to determine where the files live we can add a dependency on [`importlib_resources` and use its `files` function](https://importlib-resources.readthedocs.io/en/latest/using.html) (and we can even use pure stdlib functions in Python 3.9+).

PR #33, opened 2022-02-16
Websockets

HorstmannA:
Add initial code for simple websocket.
Websocket connection opened
Messages sent between client and server

Also: data sent in POST request


src/murfey/server/main.py
```diff
@@ -161,11 +161,25 @@ class File(BaseModel):
...
+    print("File POST received")
     return file
 
 
+@app.websocket("/ws/test")
+async def websocket_endpoint(websocket: WebSocket):
+    await websocket.accept()
+    try:
+        while True:
+            time.sleep(5)
```
Anthchirp, line 175: must not use sleep in async code

requirements_dev.txt
```diff
@@ -9,3 +9,4 @@ pytest==6.2.5
 textual==0.1.15
 uvicorn[standard]==0.17.4
 zocalo==0.14.0
+websocket-client==1.2.3
```
Anthchirp, line 12: minor nitpick: file is alphabetically sorted

Anthchirp (approved):

src/murfey/server/main.py
Anthchirp: eg. https://stackoverflow.com/a/56730924

PR #40, opened 2022-02-23
Add websockets support

HorstmannA:
Add updating functionality to websockets.
Server broadcasts to all connected clients when a new client joins, and when a post request is received.

To see this you can start a server (murfey.server), client (murfey) and send a post request eg. 
`curl -X POST http://127.0.0.1:8000/visits/cm31111-1/files --data '{"name": "microscope", "description": "8361", "size": 25, "timestamp": 24.0}' -H "Content-Type: application/json"`

### DIFF
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -6,8 +6,7 @@ import pathlib
 import threading
 
 import murfey.client.update
-from murfey.client.main import (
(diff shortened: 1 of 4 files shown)


Anthchirp:
Lgtm is right, run_forever runs in the main thread, the second thread doesn't do anything

PR #41, opened 2022-02-24
Include rsync functionality in the client run() method and take command line arguments for file transfer directories.

HorstmannA:
Include rsync which runs if 'source' and 'destination' command line arguments are provided. As files are transferred, post requests are made and the server broadcasts to connected clients.

### DIFF
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import argparse
 import configparser
 import pathlib
+import threading
 
 import murfey.client.update
 from murfey.client.main import setup_rsync, websocket_app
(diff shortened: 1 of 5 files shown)


src/murfey/client/__init__.py
```diff
@@ -16,11 +17,19 @@ def run():
...
+    args = parser.parse_args()
+    visit_name = args.visit
 
-    websocket_app()
+    ws = threading.Thread(target=websocket_app)
+    ws.start()
+    if args.source and args.destination:
+        source_directory = pathlib.Path(args.source)
+        destination_directory = pathlib.Path(args.destination)
+        setup_rsync(visit_name, source_directory, destination_directory)
```
Anthchirp, line 32: This needs to go at the end of the function, ie. after server connectivity and update checks

PR #45, opened 2022-03-07
Info from client logged by server

HorstmannA:
This is a rough first attempt and requires improvement. The file-watching part is run in the client's run() method, and notifies and logs when a new file is found. At the moment it times out after 5 minutes until we figure out how/when to stop it properly. No custom formatting is done yet, the logs just include the filename.
The logs sent to Graylog in my testing can be seen [here](https://graylog2.diamond.ac.uk/search?rangetype=relative&fields=message%2Csource&width=1920&highlightMessage=&relative=1800&q=source%3A+ws466.diamond.ac.uk).


Anthchirp (approved):
Okay, let's integrate this and clean it up later

PR #54, opened 2022-04-01
Start Data Collection in ISPyB

HorstmannA:
Note: as of this PR the server must be started with "murfey.server -e devrmq -t PikaTransport" rather than just "murfey.server" because the dev environment is no longer hardcoded.

Send messages via Zocalo to start a Data Collection.
Uses the visit given when the Murfey client is started and looks up session_id with SQLAlchemy.
Some fields in the DC entry are filled in.

### DIFF
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -130,12 +130,14 @@ def run():
         start_time = str(datetime.datetime.now())
         image_directory = str(args.destination)
         image_suffix = input("Enter the image suffix: ")
+        visit = str(args.visit)
         dc_params = {
             "type": "start_dc",
             "params": {
                 "start_time": start_time,
                 "image_directory": image_directory,
                 "image_suffix": image_suffix,
+                "visit": visit,
             },
         }
         ws.send(json.dumps(dc_params))
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -133,12 +133,10 @@ def run():
         visit = str(args.visit)
         dc_params = {
             "type": "start_dc",
-            "params": {
-                "start_time": start_time,
-                "image_directory": image_directory,
-                "image_suffix": image_suffix,
-                "visit": visit,
-            },
+            "start_time": start_time,
+            "image_directory": image_directory,
+            "image_suffix": image_suffix,
+            "visit": visit,
         }
         ws.send(json.dumps(dc_params))
 
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -125,9 +125,6 @@ def run():
     ongoing_visits = _get_visit_list(murfey_url)
     pprint(ongoing_visits)
 
-    # current_visit = [visit for visit in ongoing_visits if visit.name == args.visit]
-    # session_id = current_visit[0].session_id
-
     _enable_webbrowser_in_cygwin()
 
     log.setLevel(logging.DEBUG)
--- a/src/murfey/client/__init__.py
+++ b/src/murfey/client/__init__.py
@@ -123,7 +123,11 @@ def run():
     from pprint import pprint
 
     print("Ongoing visits:")
-    pprint(_get_visit_list(murfey_url))
+    ongoing_visits = _get_visit_list(murfey_url)
+    pprint(ongoing_visits)
+
+    current_visit = [visit for visit in ongoing_visits if visit.name == args.visit]
+    session_id = current_visit[0].session_id
 
     _enable_webbrowser_in_cygwin()
 
@@ -149,6 +153,7 @@ def run():
         dc_params = {
             "type": "start_dc",
             "start_time": start_time,
+            "session_id": session_id,
             "image_directory": image_directory,
             "image_suffix": image_suffix,
             "visit": visit,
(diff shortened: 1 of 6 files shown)


src/murfey/server/__init__.py
```diff
@@ -77,7 +79,7 @@ def filter(self, record: logging.LogRecord) -> bool:
 def run():
     # setup logging
     zc = zocalo.configuration.from_file()
-    zc.activate_environment("live")
+    zc.activate_environment("devrmq")
```
Anthchirp, line 82: we will need to expose this as an argument.

src/murfey/server/ispyb.py
```diff
@@ -21,6 +24,39 @@
...
+                sqlalchemy.func.concat(
+                    _Proposal.proposalCode,
+                    _Proposal.proposalNumber,
+                    "-",
+                    _BLSession.visit_number,
+                )
+                == visit
+            )
+        )
+        message["session_id"] = session.first()[0].sessionId
```
Anthchirp, line 55: Either extract this as a "get session for visit" if we need this - or more sensibly - if we need the session id then we should extend the `Visit` model in `util.models` (for that merge the current `main` into your branch first) to hold that extra piece of information and then there's no need to look it up specifically.

src/murfey/server/websocket.py
```diff
@@ -87,6 +91,10 @@ async def forward_log(logrecord):
     logging.getLogger(record_name).handle(logging.makeLogRecord(logrecord))
 
 
+async def start_dc(data):
+    _transport_object.start_dc(data)
```
Anthchirp, line 95: that indirection seems unnecessary

Anthchirp (approved):

src/murfey/server/__init__.py
Anthchirp: doesn't need to be _now_ now, but soon after

src/murfey/client/__init__.py
```diff
@@ -136,6 +142,24 @@ def run():
...
+        visit = str(args.visit)
+        dc_params = {
+            "type": "start_dc",
+            "start_time": start_time,
+            "session_id": session_id,
+            "image_directory": image_directory,
+            "image_suffix": image_suffix,
+            "visit": visit,
+        }
+        ws.send(json.dumps(dc_params))
```
Anthchirp, line 161: fundamentally, I think the decision when a new datacollection starts should be made by the server, not the client?

src/murfey/server/ispyb.py
```diff
@@ -58,9 +90,30 @@ def get_all_ongoing_visits(microscope: str, db: sqlalchemy.orm.Session) -> list[
...
+    query = (
+        Session()
+        .query(_DataCollectionGroup)
+        .filter(
+            _DataCollectionGroup.sessionId == session_id,
+        )
+        .all()
+    )
+    dcgids = [row.dataCollectionGroupId for row in query]
+    return dcgids
```
Anthchirp, line 119: Note: This is wasteful. The query retrieves every field and throws all of them (but one) away.

(However, the approach to inserting the DCID is more fundamentally broken so this doesn't matter much)

src/murfey/server/ispyb.py
```diff
@@ -22,6 +29,30 @@
...
+        message["ispyb_command"] = "insert_data_collection"
+        message["dcgid"] = dcgid[0]
+        ispyb_message = {"content": "Murfey DC insert", "parameters": message}
+        self.transport.send("ispyb_connector", ispyb_message)
+
+    def insert_data_collection_group(self, message):
+        message["ispyb_command"] = "insert_data_collection_group"
+        ispyb_message = {"content": "Murfey DCG insert", "parameters": message}
+        self.transport.send("ispyb_connector", ispyb_message)
+        return
```
Anthchirp, line 53: Can't do it this way, there are too many assumptions here.
Also there is no need to do it this way: you can send a [multipart message](https://github.com/DiamondLightSource/python-dlstbx/blob/950c89aa9ae0268dc722ee227cbbca36071b27ba/src/dlstbx/services/ispybsvc.py#L996) of "1. Create a DCG, 2. use the output to create a DC"
See the `ispyb-relion` recipe and how it creates, sets, and then uses `ispyb_autoprocprogram_id` for an example

src/murfey/util/models.py
```diff
@@ -17,6 +18,7 @@ def __repr__(self) -> str:
             "Visit("
             f"start='{self.start:%Y-%m-%d %H:%M}', "
             f"end='{self.end:%Y-%m-%d %H:%M}', "
+            f"session_id='{self.session_id!r}',"
```
Anthchirp, line 21:
```suggestion
            f"session_id='{self.session_id}',"
```
don't need the `repr()` call when we know it's an integer.

src/murfey/server/__init__.py
```diff
@@ -96,6 +98,11 @@ def run():
         type=int,
         default=8000,
     )
+    parser.add_argument(
+        "--transport",
+        help="Transport type for Zocalo connection (default: Pika Transport)",
+        default="PikaTransport",
+    )
```
Anthchirp, line 105: why not `workflows.transport.add_command_line_options(parser, transport_argument=True)`?

PR #66, opened 2022-08-02
Start DC tidy up

HorstmannA:
Neaten up the start_dc mechanism (but don't actually send atm)

### DIFF
--- a/.mypy.ini
+++ b/.mypy.ini
@@ -1,0 +1,5 @@
+[mypy]
+mypy_path=src
+
+[mypy-yaml.*]
+ignore_missing_imports=True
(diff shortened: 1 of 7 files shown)


d-j-hatton (approved):

HorstmannA:
Can be merged, sorry! Shall I go ahead or will it cause problems for you now? Also, shall we fix the tests which don't pass?

d-j-hatton:
It isn't causing any problems, just thought it was worth closing if possible for neatness. One problem with the tests seems to be on the Azure side (the tests on the main branch ran fine when the last commit went in but when I re-ran them they failed). The other test failures are fixed in the base branch (I think, not sure if I've remembered that correctly). This isn't going into main so I think it is fine to merge and fix tests later if needed

Issue #114, opened 2022-11-14
Check if user is sure they want to trigger processing


HorstmannA:
It's quite easy to accidentally start processing. Add a dialogue to check if they definitely want to do this, with the option to go back if not.

d-j-hatton:
Resolved by #121

PR #121, opened 2022-11-28
Check with the user that they want to proceed with processing

d-j-hatton:
After the processing parameters form is completed check with the user that they wish to continue. This gives users a chance to go back and check the parameters they entered.


d-j-hatton:
This resolves #114

PR #132, opened 2023-01-04
Use manual tilt offset from input parameters 

HorstmannA:
and pass pixel size to tomo align service.

This means the tilt offset will be taken into account by the tomo_align service, and the pixel size will be scaled and stored in the database as pixel_spacing.
This should fix https://github.com/DiamondLightSource/python-murfey/issues/129

### DIFF
--- a/src/murfey/client/context.py
+++ b/src/murfey/client/context.py
@@ -127,7 +127,8 @@ class TomographyContext(Context):
         appid: int,
         mvid: int,
         tilt_angles: List,
-        tilt_offset: Optional[float],
+        manual_tilt_offset: Optional[float],
+        pixel_size: Optional[float],
     ):
         if self._extract_tilt_series and self._extract_tilt_tag:
             tilt_series = (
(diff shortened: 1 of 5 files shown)


d-j-hatton:
I don't think it needs the changes to `instance_environment.py`. The additional parameters will end up in `data_collection_parameters` inside the `MurfeyInstanceEnvironment` which is where you get them from in `_add_tilt`

HorstmannA:
I got a bit confused with the instance_environment stuff - my thinking was that `check_for_alignment()` is called in `context.py` each time the `motion_corrected_movies` is set in `instance_environment.py`, so I need to give it all of the parameters that `check_for_alignment` needs

d-j-hatton:
Yes, you're right it does need the parameters for the callback. As we discussed these should be available from `values["data_collection_parameters"]`. This stuff is a mess and I need to clean it up somehow

d-j-hatton (approved):

PR #149, opened 2023-03-23
Add path_pattern and remove movie_id fields 

HorstmannA:
for the message sent to alignment.
I think the other movie_id references are okay to remain.


d-j-hatton:
If we leave the movie ID in will the ISPyB service still use it? I'm wondering if we want to leave it for now for safety

d-j-hatton (approved):

PR #171, opened 2023-05-19
Separate client & server in dev-portal

garryod:
Hi Dan,

Cheers for adding an entity descriptor for the developer portal! I thought it might be nice to split the entity into a client & server component to reflect the real application. I've edited the `catalog-info.yaml` file with what I think is right, but I'm not 100% sure I've captured things correctly, any chance you could answer the following?
- Is Murfey is an entire system unto itself? If so it would be good to add a `system` component representing it, if not it's probably best left out. 
- Is the client is a library or something else? It reads as if it has a web interface which I guess would make it a website under the [common component types](https://diamondlightsource.github.io/developer-portal/user/references/common-component-types/) - though if you have some other way of describing it we can expand that list.

### DIFF
--- a/catalog-info.yaml
+++ b/catalog-info.yaml
@@ -1,23 +1,53 @@
 apiVersion: backstage.io/v1alpha1
+kind: System
+metadata:
+  name: murfey
+  title: Murfey
+  description: A client-server architecture hauling Cryo-EM data around systems and triggering processing
+spec:
+  owner: user:kif41228
+
+---
+
+apiVersion: backstage.io/v1alpha1
 kind: Component
 metadata:
-  name: Murfey
-  title: Murfey
-  description: Client-Server architecture hauling Cryo-EM data around systems and triggering processing
+  name: murfey-server
+  title: Murfey Server
+  description: A server allowing for the monitoring of Cryo-EM data transfer from microscope systems and triggering relevant processing
   annotations:
     github.com/project-slug: DiamondLightSouce/python-murfey
 spec:
-  type: other
+  type: service
   lifecycle: production
   owner: user:kif41228
   providesApis:
-    - restapi
+    - murfey-rest
+
 ---
+
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: murfey-client
+  title: Murfey Client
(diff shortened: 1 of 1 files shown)


d-j-hatton:
Hi Garry,

Thanks for taking a look. I was struggling a bit to map the murfey client onto the component types. 

Murfey interacts with things like ISPyB and Zocalo but it probably is fair to describe it as a system.

The client uses [`textual`](https://github.com/textualize/textual/) so has a terminal based UI that is modelled on web technologies but is not actually a web frontend. (We're hoping to develop one once we figure out how to get around some microscope system network oddities but that will be a separate component.) Even ignoring the UI the client is still a distinct component as it has to handle data transfer. I guess `library` is the fairest description out of the common types

catalog-info.yaml
```diff
@@ -1,23 +1,53 @@
...
+  owner: user:kif41228
+
+---
+
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: murfey-server
+  title: Murfey Server
+  description: A server allowing for the hauling Cryo-EM data around systems and triggering processing
```
d-j-hatton, line 17:
```suggestion
  description: A server allowing for the monitoring of Cryo-EM data transfer from microscope systems and triggering relevant processing
```

catalog-info.yaml
```diff
@@ -1,23 +1,53 @@
...
+    - murfey-rest
+
+---
+
+apiVersion: backstage.io/v1alpha1
+kind: Component
+metadata:
+  name: murfey-client
+  title: Murfey Client
+  description: A client used to haul Cryo-EM data around systems and triggering processing
```
d-j-hatton, line 34:
```suggestion
  description: A client for automated data transfer and Cryo-EM metadata extraction from EM systems
```

garryod:
Ah, that's cool, I've never come across `textual` before. Am I right in thinking `user-interface` would be the best type to give it? If so I shall add that to the list of common types as I expect it will be used fairly extensively throughout diamond.

d-j-hatton:
Yep, `user-interface` works. At the moment there is a `tui` submodule of `client` with the interface code in it, so there is some separation there

d-j-hatton:
Cheers

d-j-hatton (approved):

PR #198, opened 2023-12-13
Fix particle diameter selection for SPA processing

d-j-hatton:
Fixes for message handling and recover the ISPyB picker ID


src/murfey/server/__init__.py
```diff
@@ -739,30 +741,61 @@ def _register_picked_particles_use_boxsize(message: dict, _db=murfey_db):
...
+        ).all()
+        for s in selection_stash:
+            _register_class_selection(
+                {
+                    "session_id": s.session_id,
+                    "class_selection_score": s.class_selection_score or 0,
+                },
+                _db=_db,
+            )
+            _db.delete(s)
```
stephen-riggs, line 771: Looks to me like there should be a _db.commit() after the delete

PR #230, opened 2024-03-11
Enable processing for EPU data directories other than `Images-Disc1`

### LINKED ISSUE #228: Processing for data directories other than `Images-Disc1`
Transfer works for other `Images-Disc` directories but processing is not started

d-j-hatton:
Hopefully harmless. Requires some testing. Resolves #228


stephen-riggs (approved):
This looks fine to me as it matches what we discussed the other day

PR #231, opened 2024-03-15
Use user-supplied tomography gain if available

stephen-riggs:
We hit an issue where the gain reference in the processing form is ignored for tomography. This will hopefully resolve that


d-j-hatton (approved):
Thanks for catching

Issue #280, opened 2024-05-21
Faster LIF to TIFF conversion


d-j-hatton:
Should be able to get a substantial speed up using multiprocessing

PR #282, opened 2024-05-21
Parallelised LIF-to-TIFF file conversion

tieneupin:
Helps with resolving #280. Part of continued work on #288.
- Added option to define number of worker threads when converting LIF files to TIFF files
- Updated logic when creating directory structure to store processed files in
- Saved metadata of LIF sub-stacks as their own XML files under each sub-stack's metadata folder

### DIFF
--- a/src/murfey/util/lif.py
+++ b/src/murfey/util/lif.py
@@ -12,7 +12,7 @@ from typing import Generator, List, Optional, Tuple
 from xml.etree import ElementTree as ET
 
 import numpy as np
-from readlif.reader import LifFile  # , LifImage
+from readlif.reader import LifFile
 from tifffile import imwrite
 
 # Create logger object to output messages with
@@ -80,7 +80,7 @@ def get_image_elements(root: ET.Element) -> List[ET.Element]:
     return elem_list
 
 
-def raise_bit_depth_error(bit_depth: int):
+def raise_BitDepthError(bit_depth: int):
     """
     Raises an exception if the bit depth value provided is not one that NumPy can
     handle.
@@ -95,14 +95,15 @@ def raise_bit_depth_error(bit_depth: int):
 
 def change_bit_depth(
     array: np.ndarray,
-    bit_depth: int,
+    target_bit_depth: int,
 ) -> np.ndarray:
     """
     Change the bit depth of the array without changing the values (barring rounding).
     """
 
-    # Use shorter terms
+    # Use shorter terms in function
     arr = array
+    bit_depth = target_bit_depth
 
     # NumPy defaults to float64; revert back to unsigned int
     if bit_depth == 8:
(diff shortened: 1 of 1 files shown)


src/murfey/util/lif.py
```diff
@@ -192,28 +184,164 @@ def rescale_to_bit_depth(
...
+
+        # Rescale intensity values for fluorescent channels
+        if any(
+            color.lower() in key
+            for key in [
+                "blue",
+                "cyan",
+                "green",
+                "magenta",
+                "red",
```
d-j-hatton, line 273: Does yellow need to be included here?

src/murfey/util/lif.py
```diff
@@ -6,13 +6,13 @@
...
 import logging
+import multiprocessing as mp
 from pathlib import Path
 from typing import Generator, List, Optional, Tuple
 from xml.etree import ElementTree as ET
 
-# import matplotlib.pyplot as plt
 import numpy as np
-from readlif.reader import LifFile
+from readlif.reader import LifFile  # , LifImage
```
d-j-hatton, line 15: Can drop commented out import

src/murfey/util/lif.py
```diff
@@ -192,28 +184,164 @@ def rescale_to_bit_depth(
...
+        # Generate slice labels
+        image_labels = [f"{f}" for f in range(len(list(image.get_iter_z())))]
+
+        # Save as a greyscale TIFF
+        save_name = img_dir.joinpath(color + ".tiff")
+        logger.info(f"Saving {color.lower()} image as {save_name}")
+        imwrite(
+            save_name,
+            arr,
+            imagej=True,  # ImageJ comppatible
```
d-j-hatton, line 317: typo on compatible, trivial but may as well catch it

src/murfey/util/lif.py
```diff
@@ -192,28 +184,164 @@ def rescale_to_bit_depth(
...
+    # Parijat wants the images in 8-bit; scale down from 16-bit
+    # Save channels as individual TIFFs
+    for c in range(len(list(image.get_iter_c()))):
+        # Get color
+        color = channels[c].attrib["LUTName"]
+        logger.info(f"Examining the {color.lower()} channel")
+
+        # Extract image data to array
+        logger.info("Loading image stack")
+        arr: np.ndarray = []  # Array to store frames in
```
d-j-hatton, line 249: This isn't an `np.ndarray` yet, it's a list

d-j-hatton: I'm surprised mypy doesn't complain about it because `np.ndarray` has no `append` method. You could make it a numpy array at initialisation and use `np.append` inside the for loop

d-j-hatton (approved):

Issue #194, opened 2023-12-08
Modernise package layout and CI


d-j-hatton:
Move to `pyproject.toml` for package description and GitHub Actions for CI

PR #296, opened 2024-05-31
Updated package dependencies

tieneupin:
Part of continued work on issue #194.

Updated Murfey's package dependencies as follows:
- Added `defusedxml` and `importlib-metadata` as explicit general dependencies (used by both client and server)
- Added `sqlalchemy[postgresql]<2` as an explicit server dependency
- Freed up version restrictions for `stomp-py` and `textual`
- Locked `numpy` to <2 while dependencies are upgrading

Additionally, updated the pre-commit hooks used to maintain the health of the repository.
- `pyproject-fmt` will be fixed at v1.8.0 for now, as in >v2.x.x, they have changed their opinion and barred the use of sub-tables for all project-related keys
- All other hooks updated to their latest versions

**Validation Checks:**
- [x] Murfey's demo server runs without issue. More extensive testing might be needed with an actual live session.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -152,7 +152,6 @@ ignore = [
 max-line-length = "88"
 
 [tool.pyproject-fmt]
-check = false
 indent = 4
 keep_full_version = true
 max_supported_python = "3.11"


d-j-hatton:
I'll do some further testing of this one. Textual updates may affect how the client UI works (or if it works)

PR #307, opened 2024-06-18
Always update the tilt series length in the database after an mdoc

stephen-riggs:
Tomogram alignment sometimes isn't requested as the database doesn't have a tilt series length. This happens if an mdoc is seen before all the tilts. To solve this, always insert tilt series length into the database when an mdoc is seen.

### DIFF
--- a/src/murfey/server/api.py
+++ b/src/murfey/server/api.py
@@ -635,6 +635,7 @@ def register_completed_tilt_series(
                     DataCollectionGroup, DataCollection, ProcessingJob, AutoProcProgram
                 )
                 .where(DataCollectionGroup.session_id == session_id)
+                .where(DataCollectionGroup.tag == tilt_series_group.source)
                 .where(DataCollection.tag == ts.tag)
                 .where(DataCollection.dcg_id == DataCollectionGroup.id)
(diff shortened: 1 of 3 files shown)


src/murfey/server/api.py
```diff
@@ -574,6 +574,32 @@ def register_tilt_series(
     db.commit()
 
 
+@router.post("/clients/{client_id}/tilt_series_length")
```
d-j-hatton, line 577: I think we can use `"/sessions/{session_id}/tilt_series_length"` to avoid the `session_id` lookup below. On the client you should be able to just use `environment.murfey_session` rather than `environment.client_id`

stephen-riggs:
There's an unpleasant logical chain I found which causes the 'multiple rows found' error which we've been seeing. The API would always get that error, but alignment normally runs as it gets triggered by motion correction not by the mdoc. That should be resolved by an extra where condition.
I also found that if alignment is triggered by motion correction it continues to use the autoprocprogram id for preprocessing rather than the alignment one.

d-j-hatton:
Thanks for tracking that stuff down

d-j-hatton (approved):

Issue #304, opened 2024-06-11
Issues with file transfer during CLEM workflow


tieneupin:
- (FIXED) Post-transfer file name of XLIF files generated incorrectly (Log: `murfey.client.contexts.tomo`).
  - Path to file incorrect; `/data/` is missing from the file path.
    - Attributed to missing `"data"` value in `"rsync_module"` key (`murfey.clients.contexts.clem`).
  - The message coming from the tomography context logger is an issue as well.

- Logic for CLEM context not working correctly (Log: `murfey.client.contexts.clem`).
   - Number of files present in series does not match that deduced from XLIF file.

- (FIXED) "Failed to transfer file" message occurs even when file is successfully transferred (Log: `murfey.client.tui.app`).
   - Related to presence of spaces in file path; conditions for handling file strings need to be tightened.
     - Attributed to `murfey.client.rsync` not parsing the stdout correctly (`class RSyncer -> def parse_stdout()`).
       - There is a `.replace(" ", "")` that messes up the file path name

- TIFF file parsing conditions need to be more flexible
  - When Leica detects a repeated Position name, it adds a `--00X` to the end of the file name.
    - E.g.: `"Position X--ZXX--CXX.tif"` and `"Position X--ZXX--CXX--XXX.tif"`.
    - This might be a non-issue, as it seems to only occur with the binned `Position X_pmd_X--...` files, which we might decide not to process.
  - Skip the transfer of "pmd" files, as they don't seem to be important for subsequent steps in the workflow and are quite numerous.
  - More robust sorting for when there are hundreds of files in a stack.

- Feedback from P: If some folders contain dud data, can we get Murfey to transfer specific folders only?

d-j-hatton:
`Analyser.post_transfer` just calls the `post_transfer` method on whatever the `Context` has been determined to be

PR #308, opened 2024-06-19
Further CLEM fixes

tieneupin:
Continued work on issues raised in #304.
- Added logic to ignore .xlif and .tif files that aren't part of the CLEM workflow
- Corrected CLEM context logic

### DIFF
--- a/src/murfey/client/contexts/clem.py
+++ b/src/murfey/client/contexts/clem.py
@@ -232,7 +232,7 @@ class CLEMContext(Context):
 
             # Post message if all files for the associated series have been collected
             # .get(series_name, 0) returns 0 if no associated key is found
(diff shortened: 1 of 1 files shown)


src/murfey/client/contexts/clem.py
```diff
@@ -221,17 +232,21 @@ def post_transfer(
 
             # Post message if all files for the associated series have been collected
             # .get(series_name, 0) returns 0 if no associated key is found
-            if len(self._tiff_series[series_name]) == 0:
+            if len(self._tiff_series.get(series_name, [])) == 0:
```
d-j-hatton, line 235:
```suggestion
            if not len(self._tiff_series.get(series_name, [])):
```
is a bit more pythonic

d-j-hatton (approved):

Issue #313, opened 2024-07-02
TIFF image stack creation in CLEM workflow can be further optimised


tieneupin:
- Metadata can be copied across more accurately and comprehensively
- (Skipped for now) Image stacks can potentially be processed in a less memory-intensive way

PR #314, opened 2024-07-03
Optimisations to TIFF image stack creation in CLEM workflow

tieneupin:
Partially resolves issue #313:
* More comprehensive transfer of metadata to created image stacks
* Improved functions used in array manipulation when creating image stacks

### DIFF
--- a/src/murfey/util/clem/images.py
+++ b/src/murfey/util/clem/images.py
@@ -13,14 +13,10 @@ from tifffile import imwrite
 # Create logger object to output messages with
 logger = logging.getLogger("murfey.util.clem.images")
 
+# Accepted bit depths and corresponding NumPy dtypes
 # For use by various functions in the script
-valid_dtypes = (
-    "uint8",
-    "uint16",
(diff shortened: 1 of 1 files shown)


src/murfey/util/clem/images.py
```diff
@@ -5,123 +5,133 @@
...
+from typing import Optional, Union
 
 import numpy as np
 from tifffile import imwrite
 
 # Create logger object to output messages with
 logger = logging.getLogger("murfey.util.clem.images")
 
+# For use by various functions in the script
+valid_dtypes = [
```
d-j-hatton, line 17: Very minor but people would often use a tuple rather than a list here as the length doesn't change

d-j-hatton:
`adjust_contrast` seems to only ever be `"stretch"`. Could it possibly take any other values?

d-j-hatton:
Not a problem, was just wondering why the switch from `bool` to `str`

d-j-hatton (approved):

PR #321, opened 2024-07-23
Add SQL databases for CLEM workflow to Murfey

tieneupin:
Part of continued development of issue #288. Also resolves issue #328.

Now that the CLEM workflow has grown in complexity, it is time to extend Murfey's database structure to record and keep track of the CLEM data collected and processed.

### DIFF
--- a/src/murfey/server/api/clem.py
+++ b/src/murfey/server/api/clem.py
@@ -218,52 +218,49 @@ def register_lif_file(
             logger.warning(traceback.format_exc())
 
     # Register child metadata if provided
-    if len(child_metadata) > 0:
-        for metadata in child_metadata:
-            try:
-                metadata_db_entry: CLEMImageMetadata = get_db_entry(
-                    db=db,
-                    table=CLEMImageMetadata,
-                    session_id=session_id,
-                    file_path=metadata,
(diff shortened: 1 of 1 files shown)


src/murfey/server/api/clem.py
```diff
@@ -0,0 +1,665 @@
...
+    # Add metadata information if provided
+    if master_metadata is not None:
+        try:
+            master_metadata = validate_and_sanitise(master_metadata)
+            clem_lif_file.master_metadata = str(master_metadata)
+        except Exception:
+            logger.warning(traceback.format_exc())
+
+    # Register child metadata if provided
+    if len(child_metadata) > 0:
```
d-j-hatton, line 221: probably more pythonic to just use 
`if child_metadata:`
here

d-j-hatton: in fact you probably don't need that at all. if the length is 0 the for loop just won't do anything

d-j-hatton:
`src/murfey/cli/add_user.py` imports from `murfey.server.auth` which is now `murfey.server.api.auth`

d-j-hatton (approved):

Issue #322, opened 2024-07-24
Migrate CLEM file processing functions to `cryoem-services`


tieneupin:
Now that the CLEM workflow has grown, it is time to move it to `cryoem-services`, which is more functionally appropriate. Murfey should primarily serve as a data hauler and message passer, without having to do any of its own processing.

PR #323, opened 2024-07-29
Migration of CLEM functions from `murfey` to `cryoem-services`

tieneupin:
Resolves issue #322 .
* Removed modules related to image stack creation functions
* Removed associated dependencies
* Removed command line entry points
* Redirected existing functions to point to new recipes


d-j-hatton (approved):
Looks good

d-j-hatton:
It is complaining about conflicts in `pyproject.toml` which should be fine to sort out

PR #325, opened 2024-07-31
When re-establishing RabbitMQ connection create new transport object

### LINKED ISSUE #324: Try and enable RabbitMQ to re-establish connection with Murfey server if it ever has to regenerate itself

d-j-hatton:
On sends from a `TransportManager` instance there is currently a check on the connection. This doesn't work as it tries to start and existing thread. A new transport object should be created and connected instead. Resolves #324

### DIFF
--- a/src/murfey/server/ispyb.py
+++ b/src/murfey/server/ispyb.py
@@ -90,13 +90,7 @@ class TransportManager:
     def send(self, queue: str, message: dict, new_connection: bool = False):
         if self.transport:
             if not self.transport.is_connected():
-                try:
-                    self.transport.disconnect()
-                except AttributeError:
-                    # If there is not _pika_thread to join on the transport
-                    # then don't need to worry about dicsonnecting
-                    pass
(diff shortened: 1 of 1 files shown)


src/murfey/server/ispyb.py
```diff
@@ -59,6 +59,17 @@ def __init__(self, transport_type):
         self.ispyb = ispyb.open()
         self._connection_callback: Callable | None = None
 
+    def reconnect(self):
+        try:
+            self.transport.disconnect()
+        except Exception:
+            log.warning(
+                "Disconnection of olf old transport failed when reconnecting",
```
stephen-riggs, line 67: Typo 'of olf old'

stephen-riggs (approved):
This looks like it should work and I can't come up with any better solutions

src/murfey/server/ispyb.py
d-j-hatton: Thanks

PR #326, opened 2024-07-31
Record tomography preprocessing in prometheus

stephen-riggs:
We aren't currently displaying the processing of tomography tilts on grafana. But we do create the prometheus metric objects needed for that, and they only require a program id.

As the program id is sent in the `motion_corrected` feedback message, I think we can just add an increment to the metric and it will work.

Do you see any reason why this wouldn't work?

### DIFF
--- a/src/murfey/server/__init__.py
+++ b/src/murfey/server/__init__.py
@@ -2438,9 +2438,7 @@ def feedback_callback(header: dict, message: dict) -> None:
                         f"No transport object found. Zocalo message would be {zocalo_message}"
                     )
 
-            prom.preprocessed_movies.labels(
-                processing_job=_pj_id(message["program_id"], murfey_db)
-            ).inc()
+            prom.preprocessed_movies.labels(processing_job=collected_ids[2].id).inc()
             murfey_db.commit()
             murfey_db.close()
             if _transport_object:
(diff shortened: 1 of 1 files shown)


d-j-hatton:
Could the processing job ID be obtained from the `collected_ids` higher up? Or is it the wrong processing job?

stephen-riggs:
> Could the processing job ID be obtained from the `collected_ids` higher up? Or is it the wrong processing job?

I'd missed that there was a processing job ID higher up. We could use that one to avoid accessing the database again, but strictly the `collected_ids` correspond to the `em-tomo-align` recipe rather than the preprocessing recipe. Which would you rather we use?

d-j-hatton:
Yeah, it depends on whether we think we would ever need the ID to be the correct one. It would be good to avoid hitting the database if we don't. The other question is what level we want it labelled at. Is it useful to know what is happening with each tilt series separately?

stephen-riggs:
I don't think it's likely to matter which ID the metric is attached to, so I'm happy to change it to not hit the database.

We probably don't need to know the processing rate for each individual tilt series, but with the current setup it's easiest to do in on a per-series level as the metrics are attached to the processing job IDs in SPA. The alternative would be to change the metric to track the data collection group ID.

d-j-hatton:
There might be some interesting stuff we can extract if we keep it as a label per tilt series so let's stick with that

stephen-riggs:
I've changed it to use an ID already taken from the database. It looks like we do have the correct processing job from earlier after all

d-j-hatton (approved):

PR #358, opened 2024-10-03
Add a tool to reinject murfey dlq messages

stephen-riggs:
Here's the outlines of a tool to pick up murfey messages from rabbitmq, and either post them to the server or reinject to rabbitmq.

It's probably functional in this form but feel free to suggest better ways to do this. Currently it is a bit clunky as it requires the token to be given then subprocess runs some zocalo commands.

I'd suggest we leave this as an open PR until I've had the opportunity to test this on a real case.


d-j-hatton (approved):
We might be able to do something better with the token at some point, like prompt for username and password to generate token and then just use it internally but fine for now

Issue #318, opened 2024-07-16
`rsync` flags some files as failing to transfer even though the file was successfully transferred


tieneupin:
These appear to be false positives due to the file path changing between when the file is first detected and after it's transferred. The file is present at the destination location even though the error was triggered. Will need to figure out what changed, and how the `rsync` module can be optimised to prevent this from occurring further.

PR #360, opened 2024-10-09
More `rsync` fixes

tieneupin:
A follow-up to the rsync issue raised in #318 .
Replaced the individual file warnings in `murfey.client.rsync` with a summary message comparing the files to be transferred and the files registered as successfully transferred. This should help pinpoint whether it's a file path mismatch issue, or another condition further upstream that is not behaving as expected.

Additionally, the MSYS2 environment is set up differently from Cygwin, so Python `pathlib.Path`  defaults to a `WindowsPath` object instead of a `PosixPath` as in Cygwin. This PR also introduces additional conditions to ensure file paths are passed to the rsync daemon in the correct Unix format.

### DIFF
--- a/src/murfey/util/__init__.py
+++ b/src/murfey/util/__init__.py
@@ -87,6 +87,9 @@ def posix_path(path: Path) -> str:
     Converts a Windows-style path into a Posix one. Used primarily when running
     subproceses in bash terminals on Windows devices, which can only accept
     Posix paths.
+
+    Returns it as a string because this path won't be recognised as an existing
+    path when converted to a Path object.
     """
     path_parts = list(path.parts)
     # Check if it's a Windows-style path and converts it to a Posix one


d-j-hatton (dismissed):

src/murfey/client/tui/screens.py
```diff
@@ -906,18 +906,22 @@ def on_button_pressed(self, event):
...
-            cmd = [
+            # Set up rsync command
+            rsync_cmd = [
                 "rsync",
-                str(self._dir_tree._gain_reference),
+                f"{posix_path(self._dir_tree._gain_reference)!r}",
                 f"{self.app._environment.url.hostname}::{visit_path}/processing/{secure_filename(self._dir_tree._gain_reference.name)}",
             ]
+            # Encase in bash shell
+            cmd = ["bash", "-c" " ".join(rsync_cmd)]
```
stephen-riggs, line 916: missing comma after -c

src/murfey/util/__init__.py
```diff
@@ -82,6 +82,22 @@ def secure_path(in_path: Path, keep_spaces: bool = False) -> Path:
     return Path("/".join(secured_parts))
 
 
+def posix_path(path: Path) -> str:
```
stephen-riggs, line 85: why does this return str rather than Path?

stephen-riggs (approved):
One bug and one query, otherwise ready to merge

PR #376, opened 2024-10-18
Remove calls to get tilt series and tag

stephen-riggs:
Removes the functions to extract the tilt series and tag from tomography files, and instead uses the existing `_construct_tilt_series_name` method.

All the tests passed, but it's likely this removes back-compatibility with older tomo versions and SerialEM.

Also changes the angle index determination to be match the function in `cryoem-services`


d-j-hatton (approved):
If the linters didn't pick up any stray references to the removed attributes then seems fine, thanks

PR #377, opened 2024-10-22
Session creation post was using the wrong endpoint

stephen-riggs:


stephen-riggs:
That was strange: the tests initially failed because they tried to run while the repository with the sql runner dependency was in the process of making a new release

d-j-hatton (approved):

PR #379, opened 2024-10-24
Remove the need for an ISPyB instance

d-j-hatton:
ISPyB may not be in use at other facilities. This removes the requirement for a connection to an ISPyB database. The option to manually name visits is provided and data collection IDs etc. are autoincremented within the Murfey database when there is no ISPyB to interact with.


stephen-riggs (approved):
I can't see any obvious issues

PR #387, opened 2024-10-25
Murfey database documentation

tieneupin:
Updated the package READMEs to include latest packages used and instructions on setting up the PostgreSQL database for use with Murfey.


d-j-hatton:
If you rename the `MURFEY_DB.md` file to `README.md` it should display in GitHub

src/murfey/server/MURFEY_DB.md
```diff
@@ -0,0 +1,84 @@
+# Murfey Database
+
+Murfey has been designed to work with PostgreSQL databases to maintain a record of the files it has transferred, which allow it to oversee and manage multi-step data processing workflows. This page will walk readers through the process of setting up a PostgreSQL database for use with Murfey.
+
+## Setting Up
+
+### Installing `postgresql`
+
+To start, we will install the `postgresql` Python package hosted on conda-forge. This can be done using either `conda` or `mamba`. In this example, we will create a new Python 3.9 environment named `murfey-db` which will be used to run `postgresql`.
```
d-j-hatton, line 9: it isn't a Python package, so I'd just say "package"

src/murfey/server/MURFEY_DB.md
```diff
@@ -0,0 +1,84 @@
...
+
+```text
+(base)$ mamba create -n murfey-db python==3.9.*
+(base)$ mamba activate murfey-db
+(murfey-db)$ mamba install -c conda-forge postgresql
+```
+
+### Creating and initialising the database
+
+Next, we need to find a directory for the database to reside in. This can generally be any location in your file system, but for this example, with a Unix OS as reference, we shall create a directory called `murfey` in the default home location of `~`.
```
d-j-hatton, line 19: can probably drop the bit about Unix OS, we almost certainly aren't Windows compatible so wouldn't worry about anyone running there

d-j-hatton:
Thanks, looks good. We should discuss how to make the Murfey configuration setup easier with a few command line tools.

src/murfey/server/MURFEY_DB.md
```diff
@@ -0,0 +1,84 @@
...
+Next, we will need to create a role in the Murfey database under whose name the data will be registered. To do so, we start up `postgresql` and run the following commands:
+
+```text
+(murfey-db)$ psql murfey -U murfey -W
+murfey=# create user murfey_server with password '[password here]' createdb;
+murfey=# grant all privileges on database murfey to murfey_server;
+murfey=# grant all privileges on schema public to murfey_server;
+```
+
+The user 'murfey_server' will be stored under the 'username' key in the database credentials file used for this microscope, and the database name 'murfey' is stored under the 'database' key in that same file. The password you use when creating this role should be the encrypted one set up when creating the credentials file.
```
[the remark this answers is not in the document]
d-j-hatton: sorry, yes, read it wrong

d-j-hatton (approved):

PR #392, opened 2024-10-31
Make machine configuration extendable

d-j-hatton:
Allow extra configuration options to be extracted from the Murfey machine configuration and validated against a pydantic model specified as an entry point. This allows external packages to add their configuration to the Murfey config for convenience

### DIFF
--- a/src/murfey/util/config.py
+++ b/src/murfey/util/config.py
@@ -164,7 +164,7 @@ def get_extended_machine_config(
     extension_name: str, instrument_name: str = ""
 ) -> Optional[BaseModel]:
     machine_config = get_machine_config(instrument_name=instrument_name).get(
-        instrument_name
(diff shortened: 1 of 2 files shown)


src/murfey/util/config.py
```diff
@@ -157,3 +158,16 @@ def get_machine_config(instrument_name: str = "") -> Dict[str, MachineConfig]:
             Path(settings.murfey_machine_configuration), microscope
         )
     return machine_config
+
+
+def get_extended_machine_config(
+    extension_name: str, instrument_name: str = ""
+) -> Optional[BaseModel]:
```
stephen-riggs, line 165: The return type of this isn't directly a BaseModel

stephen-riggs (approved):

PR #395, opened 2024-11-01
Fix use of entry points

d-j-hatton:


stephen-riggs (approved):

d-j-hatton:
> Might be worth fixing other occurrences of `entry_points` in the repository.
> 
> If `backports.entry_points_selectable` is the preferred way of importing `entry_points` going forwards, we might want to phase out instances of `entry_points` imported from `importlib_metadata` as well.

True, but might do it else where to keep this one simple

PR #396, opened 2024-11-01
Added workflows to register CLEM preprocessing results

tieneupin:
Part of ongoing work for #288 . Murfey can now identify messages related to the CLEM preprocessing workflow and register the outputs in the database accordingly.

### DIFF
--- a/src/murfey/server/api/clem.py
+++ b/src/murfey/server/api/clem.py
@@ -1,12 +1,13 @@
 from __future__ import annotations
 
 import re
-import sys
 import traceback
+from importlib.metadata import EntryPoint  # type hinting only
 from logging import getLogger
 from pathlib import Path
 from typing import Optional, Type, Union
 
+from backports.entry_points_selectable import entry_points
 from fastapi import APIRouter
 from sqlalchemy.exc import NoResultFound
 from sqlmodel import Session, select
@@ -23,12 +24,6 @@ from murfey.util.db import (
 )
 from murfey.util.db import Session as MurfeySession
 from murfey.util.models import TIFFSeriesInfo
-
-# Use backport from importlib_metadata for Python <3.10
-if sys.version_info.major == 3 and sys.version_info.minor < 10:
-    from importlib_metadata import EntryPoint, entry_points
-else:
-    from importlib.metadata import EntryPoint, entry_points
 
 # Set up logger
 logger = getLogger("murfey.server.api.clem")


src/murfey/server/__init__.py
```diff
@@ -2964,6 +2971,29 @@ def feedback_callback(header: dict, message: dict) -> None:
...
+            "register_tiff_preprocessing_result",
+        ):
+            murfey_workflows = list(
+                entry_points().select(
+                    group="murfey.workflows.clem", name=message["register"]
+                )
+            )
+            # Run the workflow if a match is found
+            if len(murfey_workflows) > 0:
+                workflow: EntryPoint = murfey_workflows[0]
```
stephen-riggs, line 2985: This should probably loop over all found entry points rather than just running the first one

pyproject.toml
```diff
@@ -97,9 +97,11 @@ murfey = "murfey.client:run"
 "password" = "murfey.server.api.auth:password_token_validation"
 [project.entry-points."murfey.config.extraction"]
 "murfey_machine" = "murfey.util.config:get_extended_machine_config"
-[project.entry-points."murfey.workflows"]
-"lif_to_stack" = "murfey.workflows.lif_to_stack:zocalo_cluster_request"
-"tiff_to_stack" = "murfey.workflows.tiff_to_stack:zocalo_cluster_request"
+[project.entry-points."murfey.workflows.clem"]
```
stephen-riggs, line 100: I wonder if we want the ".clem" bit in the name. We probably do for now so it's clear what is happening, but it might be more general to remove that when we migrate other parts of the code?

src/murfey/server/api/clem.py
```diff
@@ -631,51 +631,79 @@ def register_image_stack(
...
+        group="murfey.workflows.clem", name="process_raw_lifs"
     )
 
     # Use entry point if found
-    if len(murfey_workflows) == 1:
+    if murfey_workflows:
+
+        # Get instrument name from the database
+        # This is needed in order to load the correct config file
+        row: MurfeySession = db.exec(
```
stephen-riggs, line 652: "row" as a variable name might be a bit unclear. "session_row"?

src/murfey/workflows/clem/register_preprocessing_results.py
```diff
@@ -0,0 +1,397 @@
...
+        db.exec(select(MurfeySession).where(MurfeySession.id == session_id))
+        .one()
+        .instrument_name
+    )
+    machine_config = get_machine_config(instrument_name=instrument_name)[
+        instrument_name
+    ]
+    rsync_basepath = machine_config.rsync_basepath
+    try:
+        base_path = list(rsync_basepath.parents)[-2].as_posix()
```
stephen-riggs, line 74: Not sure I understand why you're finding the parents of the rsync_basepath? Shouldn't the file always be relative to the rsync_basepath?

src/murfey/workflows/clem/register_preprocessing_results.py
```diff
@@ -0,0 +1,397 @@
...
+        )
+
+    # Check that full file path doesn't contain unallowed characters
+    # Currently allows only:
+    # - words (alphanumerics and "_"; \w),
+    # - spaces (\s),
+    # - periods,
+    # - dashes,
+    # - forward slashes ("/")
+    if bool(re.fullmatch(r"^[\w\s\.\-/]+$", str(full_path))) is False:
```
stephen-riggs, line 90: Is this required? Surely if it passes the "is_file()" check below then it is of a valid format?

src/murfey/server/__init__.py
stephen-riggs: Sorry, I misread what it was doing slightly and thought it was checking all the "murfey.workflows.clem" entry points. You're right it will only find one entry so fine to just run that one.
Up to you whether you want to search through the other clem ones too

src/murfey/server/__init__.py
```diff
@@ -61,6 +62,12 @@
 from murfey.util.spa_params import default_spa_parameters
 from murfey.util.state import global_state
 
+# Import entry_points depending on Python version
+if sys.version_info < (3, 10):
+    from importlib_metadata import EntryPoint, entry_points
+else:
+    from importlib.metadata import EntryPoint, entry_points
+
```
d-j-hatton, line 70: can use `backports` instead

d-j-hatton (approved):

PR #400, opened 2024-11-08
Replaced 'entry_points' instances with 'backports.entry_points_selectable' one

tieneupin:
Also added comments to note when external entry points are being looked for.


d-j-hatton (approved):
We'll have to remember to install backports on the clients

PR #401, opened 2024-11-11
Setup CLI

tieneupin:
Adds a command line tool intended to walk non-developers through the process of setting up a configuration file in order to run Murfey properly.

**Key Changes:**
- The `Security` class has been renamed to `GlobalConfig`, and functions referring to it have likewise been renamed.
  - Murfey will likewise now look for the `MURFEY_GLOBAL_CONFIGURATION` environment variable instead of `MURFEY_SECURITY_CONFIGURATION`.
- Many `MachineConfig` keys have been re-typed to take `Optional[Path]` instead of `str` as part of making more parts of the Murfey workflow optional for other users. The code parsing logic has been changed to reflect this.


d-j-hatton:
In the tests the environment variables need setting back to `MURFEY_SECURITY_CONFIGURATION`

d-j-hatton:
Could we leave `(press Enter to leave it blank as '')` out for non-strings

d-j-hatton:
If nothing is added to something with a default can we give the option to choose to use the default?

d-j-hatton:
Can we have `(y/n)` in a different colour to make it clear you should answer with a yes or no?

d-j-hatton:
The recipe keys can be a literal

d-j-hatton:
If a section is disabled then the rest of the keys for the section can be skipped

PR #420, opened 2024-12-12
Added FastAPI endpoint to trigger image alignment and merging job for CLEM workflow and updated function parameters

tieneupin:
Also migrates the Pydantic model associated with the TIFF file processing workflow to the CLEM API page, to keep supporting objects localised.

This PR will need to be deployed alongside [this PR](https://github.com/DiamondLightSource/cryoem-services/pull/91) due to breaking changes.

### DIFF
--- a/src/murfey/workflows/clem/register_preprocessing_results.py
+++ b/src/murfey/workflows/clem/register_preprocessing_results.py
@@ -12,7 +12,6 @@ import logging
 import traceback
 from ast import literal_eval
 from pathlib import Path
-from typing import Literal, Optional
 
 from pydantic import BaseModel, validator
 from sqlmodel import Session, select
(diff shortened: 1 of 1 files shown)


src/murfey/server/api/clem.py
```diff
@@ -687,3 +694,72 @@ def process_raw_tiffs(
...
     return True
+
+
+class AlignAndMergeParams(BaseModel):
+    # Processing parameters
+    series_name: str
+    images: list[Path]
+    metadata: Path
+    # Optional processing parameters
+    crop_to_n_frames: Optional[int] = None
```
stephen-riggs, line 705: As noted on the cryoemservices PR, these could possibly default to something else rather than None

src/murfey/workflows/clem/align_and_merge.py
```diff
@@ -25,6 +25,7 @@ def submit_cluster_request(
     images: list[Path],
     metadata: Path,
     # Optional processing parameters
+    crop_to_n_frames: Optional[int] = None,
```
stephen-riggs, line 28: As above. Something to consider

d-j-hatton (approved):

PR #434, opened 2025-01-09
Load RabbitMQ credentials directly using PikaTransport object

tieneupin:
Now that we've removed the Zocalo configuration dependency, we import and use PikaTransport directly to load the RabbitMQ configuration

### DIFF
--- a/src/murfey/server/__init__.py
+++ b/src/murfey/server/__init__.py
@@ -11,7 +11,7 @@ from functools import partial, singledispatch
 from importlib.resources import files
 from pathlib import Path
 from threading import Thread
-from typing import Any, Dict, List, NamedTuple, Tuple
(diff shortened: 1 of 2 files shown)


src/murfey/server/__init__.py
```diff
@@ -273,28 +274,29 @@ def run():
...
-
-    # Set up Zocalo connection
     if args.demo:
+        # Run in demo mode with no connections set up
         os.environ["MURFEY_DEMO"] = "1"
     else:
-        _set_up_transport(args.transport)
+        # Load RabbitMQ configuration and set up the connection
+        PikaTransport().load_configuration_file(security_config.rabbitmq_credentials)
+        _set_up_transport("pika")
```
d-j-hatton, line 299: I think this should be `_set_up_transport("PikaTransport")`

d-j-hatton (approved):

PR #454, opened 2025-01-20
Register missed grid squares and foil holes during SPA processing flush

stephen-riggs:
Currently, if a grid square is seen before the data collection group is made, we get an error.
This addresses that by skipping the registration if there is not data collection group, and instead registering any missed grid squares and foil holes when the flush happens.

There's not much really "new" code in here, but a lot of functions have been moved and modified to allow the registration to happen from a client API call or a server call.

- Functions from `client/contexts/spa.py` are now needed by the server, so have moved into the `util` folder (with the exceptions of `_grid_square_metadata_file` which needs to behave differently on the server)
- Parts of the `api/__init__.py` are now called through the server, so have moved into a workflow.

The `flush_spa_processing` call has now moved to a separate workflow. This workflow is also the location of the additions in this PR: `flush_spa_processing` now calls a new function `_flush_position_analysis` which fills in any missing grid squares and foil holes in the database.


src/murfey/workflows/spa/flush_spa_preprocess.py
```diff
@@ -0,0 +1,391 @@
...
+            try:
+                foil_hole_id = _flush_position_analysis(
+                    movie_path=f.file_path,
+                    dcg_id=collected_ids[0].id,
+                    session_id=session_id,
+                    db=db,
+                )
+            except Exception as e:
+                logger.error(
+                    f"Flushing position analysis for {f.file_path} caused exception {e}"
```
d-j-hatton, line 337:
```suggestion
                    f"Flushing position analysis for {f.file_path} caused exception {e}", exc_info=True
```

d-j-hatton (approved):

PR #455, opened 2025-01-21
Register all grid squares and foil holes

d-j-hatton:
Register all grid squares and foil holes in `.dm` files when they are seen. For the grid squares this is the only time this information is available so needs to be done.

### DIFF
--- a/src/murfey/client/contexts/spa_metadata.py
+++ b/src/murfey/client/contexts/spa_metadata.py
@@ -6,20 +6,17 @@ import requests
 import xmltodict
 
 from murfey.client.context import Context
-from murfey.client.contexts.spa import (
-    FoilHole,
-    _get_grid_square_atlas_positions,
-    _get_source,
-)
+from murfey.client.contexts.spa import _get_source
 from murfey.client.instance_environment import MurfeyInstanceEnvironment, SampleInfo
 from murfey.util import authorised_requests, capture_post, get_machine_config_client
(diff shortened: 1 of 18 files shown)


src/murfey/client/contexts/spa_metadata.py
```diff
@@ -15,6 +15,71 @@
 requests.get, requests.post, requests.put, requests.delete = authorised_requests()
 
 
+class FoilHole(NamedTuple):
```
stephen-riggs, line 18: Can you reuse the class which was in the client, and gets moved to util in my PR? The models are very similar, but the other one has foil hole and grid square IDs

src/murfey/client/contexts/spa_metadata.py
```diff
@@ -108,36 +173,47 @@ def post_transfer(
...
+                                "height": pos_data[5],
+                                "angle": pos_data[6],
+                            },
+                        )
+
+        elif transferred_file.suffix == ".dm" and environment:
+            gs_name = transferred_file.name.split("_")[1]
+            fh_positions = _foil_hole_positions(transferred_file, int(gs_name))
+            source = _get_source(transferred_file, environment=environment)
+            visitless_source = str(source).replace(f"/{environment.visit}", "")
```
stephen-riggs, line 200: This is probably the best we can do, but it would break if the metadata isn't in the same base directory as the data.
e.g. the situation we had on the Talos with metadata in DoseFractions and data in DoseFractions/DATA

stephen-riggs (approved):

PR #469, opened 2025-01-27
Need frame counts for tomography dose weighting

stephen-riggs:
Adds frame count finding for tomography, and passes this on to the processing.
Storing the frame count requires a database change, so we may not want to merge this change immediately.


src/murfey/client/contexts/tomo.py
```diff
@@ -682,6 +689,7 @@ def gather_metadata(
             mdoc_metadata: OrderedDict = OrderedDict({})
             mdoc_metadata["experiment_type"] = "tomography"
             mdoc_metadata["voltage"] = float(mdoc_data["Voltage"])
+            mdoc_metadata["frame_count"] = float(mdoc_data_block["NumSubFrames"])
```
d-j-hatton, line 692: int?

PR #470, opened 2025-01-28
Ensure foil hole id is in the database before inserting movie

stephen-riggs:
There is a condition we have hit on the flush where movies cannot be inserted as the foil hole is not inserted. 

This will hopefully ensure that flushed movies only have a foil hole if it is present in the murfey database.


d-j-hatton:
Do you know why this is happening? Seems like a case we just never want to happen

stephen-riggs:
I think the third commit on here is probably the answer: it's trying to use the foil hole EPU id, not the murfey db id.
But I'm not sure, so I've tried to catch every case here

d-j-hatton (approved):

PR #492, opened 2025-02-12
Data directory name got duplicated when processing

stephen-riggs:


src/murfey/client/tui/screens.py
```diff
@@ -92,10 +92,7 @@ def determine_default_destination(
     elif machine_data.get("data_directories"):
         for data_dir in machine_data["data_directories"]:
             if source.resolve() == Path(data_dir):
-                _default = (
-                    destination
-                    + f"{machine_data.get('rsync_module') or 'data'}/{visit}"
-                )
+                _default = destination + str(visit)
```
d-j-hatton, line 95: I think `visit` must already be a string

d-j-hatton (approved):

PR #503, opened 2025-03-05
Container setup for TEM workflow + tomography/SPA bugfixes

tieneupin:
This PR sets up the Helm charts needed in order to create an instrument server container to test the TEM workflow with. Along the way, various bugs that were uncovered were also patched:

* Added an `rsync_url` key to the `MurfeyInstanceEnvironment` class, now that we are supporting externally run rsync daemons
* Set microscope name using `instrument_name` from database lookup, instead of using the `get_microscope()` function
* Added logs to the `prepare_(eer_)gain()` function to see why the subprocess fails
* Fixed logic for why the first tomography position is not processed
* Added logic to handle cases in the `feedback_callback` function where no workflows are found
* Locked NumPy to <2 (NumPy compatibility will need to be addressed in future; see #508 )


src/murfey/server/__init__.py
```diff
@@ -2942,12 +2940,17 @@ def feedback_callback(header: dict, message: dict) -> None:
...
+            # Search for corresponding workflow
+            workflows: list[EntryPoint] = list(
                 entry_points().select(
                     group="murfey.workflows", name=message["register"]
                 )
-            )[0]
+            )  # Returns either 1 item or empty list
+            if not workflows:
+                logger.error(f"No workflow found for {sanitise(message['register'])}")
+                return None
```
stephen-riggs, line 2951: You'll need to nack the message before returning

stephen-riggs (approved):

PR #535, opened 2025-03-24
Added FastAPI endpoints for Rust package repository

tieneupin:
Added some FastAPI endpoints to serve Rust configs, metadata, and crates from the official `crates.io` website.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -54,8 +54,7 @@ developer = [
 ]
 instrument-server = [
     "fastapi[standard]",
-    "python-jose[cryptography]",
-    "uvicorn[standard]",
+    "python-jose",
 ]
 server = [
     "aiohttp",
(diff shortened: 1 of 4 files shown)


src/murfey/server/api/bootstrap.py
```diff
@@ -565,6 +570,388 @@ def get_msys2_package_file(
...
+    different from the standard path convention. They are stored under
+    /1/{package} or /2/{package}.
+    """
+
+    logger.debug(f"Received request to access {str(request.url)}")
+
+    # Validate path to crate
+    if n not in ("1", "2"):
+        raise ValueError("Invalid path to package metadata")
+    if not bool(re.fullmatch(r"[\w\-]{1,2}", package)):
```
d-j-hatton, line 728: I wouldn't expect this `bool` cast to be necessary. You could do a `if  re.fullmatch(r"[\w\-]{1,2}", package) is None` to be safe

d-j-hatton (approved):

Issue #530, opened 2025-03-21
Finalising all rsyncers through the instrument server seems to cause connection issues with the database


d-j-hatton:
The first error log:

`sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.

[SQL: SELECT session.id, session.name, session.visit, session.started, session.current_gain_ref, session.instrument_name 
FROM session 
WHERE session.id = %(id_1)s]
[parameters: {'id_1': 22}]
(Background on this error at: https://sqlalche.me/e/20/e3q8)`

occurring in `validate_instrument_server_session_token` in auth. 

The rsyncer finalisatons happen outside of threads (https://github.com/DiamondLightSource/python-murfey/blob/7e412424845eda96abed07165b781f433cc6ad56/src/murfey/client/multigrid_control.py#L169) which may cause issues

d-j-hatton:
I think ignore the `"thread": False` thing. That's because it is already happening in a thread

PR #543, opened 2025-03-27
Allow visit completion to run more asynchronously

d-j-hatton:
The multigrid controller should now be able to handle most of the process of cleaning up all associated threads. Thread completion requests are performed by setting flags and not explicitly waiting for `join`s. Callbacks are then used to notify of thread completion. (`RSyncer` threads are slightly more complicated because they have to perform file removal after completing.) 

On completion of all associated threads the multigrid controller will mark itself as dormant. When new multigrid controllers are added dormant controllers are checked for and removed along with the associated Murfey session data. This could also possibly be triggered when the controller marks itself as dormant for a quicker cleanup?

### DIFF
--- a/src/murfey/instrument_server/api.py
+++ b/src/murfey/instrument_server/api.py
@@ -179,7 +179,9 @@ def start_multigrid_watcher(
             destination_overrides=watcher_spec.destination_overrides,
         )
     )
-    watchers[session_id].subscribe(controllers[session_id]._multigrid_watcher_finalised)
+    watchers[session_id].subscribe(
+        controllers[session_id]._multigrid_watcher_finalised, final=True
+    )
     watchers[session_id].start()
     return {"success": True}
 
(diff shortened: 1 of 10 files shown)


d-j-hatton:
Attempts to resolve #530

src/murfey/instrument_server/api.py
```diff
@@ -144,6 +144,13 @@ def start_multigrid_watcher(
     if controllers.get(session_id) is not None:
         return {"success": True}
     label = watcher_spec.label
+    for sid, controller in controllers.items():
+        if controller.dormant:
+            del controllers[sid]
+            requests.delete(
+                f"{_get_murfey_url()}/sessions/{sanitise_nonpath(str(sid))}",
+                headers={"Authorization": f"Bearer {tokens[sid]}"},
+            )
```
d-j-hatton, line 153: switch these

d-j-hatton:
Need to check that controller is dormant before deleting session from database

Issue #446, opened 2025-01-13
Client-side dependencies now also needed on server side


tieneupin:
Encountered these issues while trying to run `murfey.server --demo` after installing only `murfey[server]`.
- The client-side `.murfey` file is needed
- Textual is needed
```
$ murfey.server --demo

Murfey client configuration file /root/.murfey not found
Traceback (most recent call last):
  File "/venv/bin/murfey.server", line 5, in <module>
    from murfey.server import run
  File "/venv/lib/python3.12/site-packages/murfey/server/__init__.py", line 51, in <module>
    from murfey.client.contexts.tomo import _midpoint
  File "/venv/lib/python3.12/site-packages/murfey/client/__init__.py", line 26, in <module>
    import murfey.client.rsync
  File "/venv/lib/python3.12/site-packages/murfey/client/rsync.py", line 21, in <module>
    from murfey.client.tui.status_bar import StatusBar
  File "/venv/lib/python3.12/site-packages/murfey/client/tui/status_bar.py", line 18, in <module>
    from textual.reactive import Reactive
ModuleNotFoundError: No module named 'textual'
```

PR #563, opened 2025-04-08
Migrate client-side utility functions to a separate module

tieneupin:
Fully resolves issue #446 by migrating the client-side utility functions into a separate module named `murfey.util.client`, thereby preventing unwanted import chains. Also adds unit tests for the migrated `read_config()` and `_get_visit_list()` functions.

Additionally, added some logic into `read_config()` to allow us to specify config files directly using `MURFEY_CLIENT_CONFIGURATION` (same naming pattern as `MURFEY_MACHINE_CONFIGURATION` and `MURFEY_SECURITY_CONFIGURATION`) so that the file name doesn't have to default to `.murfey`.

**Verification Checks**
- [x] Server-side commands such as `murfey.create_db` and `murfey.add_user` no longer throw the "Client config not found" warning
- [x] TUI workflows run
  - [x] SPA
  - [x] Tomo
  - [x] CLEM
- [x] Web UI workflows run
  - [x] SPA
  - [x] Tomo
  - [x] CLEM

### DIFF
--- a/src/murfey/client/analyser.py
+++ b/src/murfey/client/analyser.py
@@ -22,8 +22,7 @@ from murfey.client.contexts.tomo import TomographyContext
 from murfey.client.instance_environment import MurfeyInstanceEnvironment
 from murfey.client.rsync import RSyncerUpdate, TransferResult
 from murfey.client.tui.forms import FormDependency
-from murfey.util import Observer
-from murfey.util.client import get_machine_config_client
+from murfey.util.client import Observer, get_machine_config_client
 from murfey.util.mdoc import get_block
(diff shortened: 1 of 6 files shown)


src/murfey/client/analyser.py
```diff
@@ -22,7 +22,8 @@
 from murfey.client.instance_environment import MurfeyInstanceEnvironment
 from murfey.client.rsync import RSyncerUpdate, TransferResult
 from murfey.client.tui.forms import FormDependency
-from murfey.util import Observer, get_machine_config_client
+from murfey.util import Observer
```
stephen-riggs, line 25: The Observer is also client-side only so could move into util/client.py

Issue #587, opened 2025-05-06
ISPyB credentials still need to be passed in as an environment variable to load the ISPyB database correctly


tieneupin:
While trying to run a Murfey server locally, I noticed that the existing visits weren't displaying correctly. Upon further investigation, I have narrowed it down to this code block:
https://github.com/DiamondLightSource/python-murfey/blob/368dd2fe087bd9b026db3780275d4cddd55a5db4/src/murfey/server/ispyb.py#L40-L45

When the `ISPYB_CREDENTIALS` environment variable is not set, the database fails to load. It looks like the ISPyB credentials that we are providing as part of the security configuration file aren't being taken into account.

Ideally, this request to load the database should make use of the ISPyB credentials provided in the security configuration file.

PR #588, opened 2025-05-06
Remove need for 'ISPYB_CREDENTIALS' environment variable

tieneupin:
Resolves issue #587 . 

`ispyb.sqlalchemy.url()` takes `credentials` as an argument, through which the config file can be passed directly to the function instead of relying on the `ISPYB_CREDENTIALS` environment variable. This eliminates the last traces of our repo's need to explicitly define `ISPYB_CREDENTIALS`.

Additionally, elements of the CI workflow and test code related to the Murfey and ISPyB databases have also been modified. `ISPYB_CREDENTIALS` has also been dropped successfully from the test workflow, and fixtures have been put in place to allow for the creation of databases that will be reset to a desired initial state at the end of every test. This means that the database does not have to be shut down and rebooted with every database-related test, which should help with our testing times as we continue to improve Murfey's test coverage.

NOTE: Within our code base, it looks like ISPyB spawns SQLAlchemy `Session` objects, whereas the Murfey database spawns SQLModel ones. As such, SQLModel's functions might not work when querying the ISPyB database. The database session fixtures for the ISPyB and Murfey test databases have been written to reflect this.

### DIFF
--- a/src/murfey/cli/spa_ispyb_messages.py
+++ b/src/murfey/cli/spa_ispyb_messages.py
@@ -19,7 +19,7 @@ from sqlmodel import create_engine, select
 
 from murfey.client.contexts.spa import _get_xml_list_index
 from murfey.server import _murfey_id, _register
-from murfey.server.ispyb import Session, TransportManager, get_session_id
+from murfey.server.ispyb import ISPyBSession, TransportManager, get_session_id
 from murfey.server.murfey_db import url
 from murfey.util import db
 from murfey.util.config import get_machine_config, get_microscope, get_security_config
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         python-version: ["3.9", "3.10", "3.11"]
     services:
       mariadb:
-        image: mariadb:11.7.2 # released 2024-05-06
+        image: mariadb:10.6 # released 2024-05-06
```
d-j-hatton, line 20: Was an auto update of this causing problems?

d-j-hatton (approved):

PR #604, opened 2025-06-10
TUI authentication fix

tieneupin:
After implementing proper authentication between the backend, the client, and the web frontend, the TUI's previous authentication method was not considered, leading to it failing authentication when connecting to the backend server after the recent update.

This PR fixes that by adding an additional `elif ...` block that takes the TUI's token into account and authorises that. In the longer term, having the TUI use the same authentication method as the instrument server would be ideal, but this is a lower priority.


d-j-hatton:
I was thinking maybe we should have a config option to turn this behaviour off so we can make some systems more secure

Issue #624, opened 2025-07-08
The migration to Pydantic 2 causes a permissions error when Murfey tries to auto-update itself


tieneupin:
It seems to be tied to Pydantic v2's use of compiled files which were not present in Pydantic v1. By the time the auto-update function is called, these files will have been imported in the process, locking the package and preventing it from being overwritten by the auto-update subprocess.

**Checks:**
* Auto-update from 0.18.2 to 0.19.1 works
* Auto-update from 0.19.0 to 0.19.1 doesn't

PR #625, opened 2025-07-08
Fixed Murfey's Auto-Update Function and Improved Pytest Functionality

tieneupin:
Resolves issue #624 .

The upgrade to Pydantic v2 has caused our current method of getting Murfey to check for updates on the client side to fail, as the files become locked and cannot be overwritten as part of the install. 

Avoiding importing Murfey components that use Pydantic appears to resolve the issue, so the `murfey.client` modules have been refactored so that Pydantic-related components are not loaded when starting the TUI or instrument server until after the update check.

Additionally, this PR also adds tests for the client update and instrument server initialisation functions, and adds functionality to Pytest to skip database-related tests if it is being run outside of GitHub, allowing us to run it locally even if we don't have the database components set up.

### DIFF
--- a/src/murfey/client/analyser.py
+++ b/src/murfey/client/analyser.py
@@ -19,6 +19,7 @@ from murfey.client.contexts.clem import CLEMContext
 from murfey.client.contexts.spa import SPAModularContext
 from murfey.client.contexts.spa_metadata import SPAMetadataContext
 from murfey.client.contexts.tomo import TomographyContext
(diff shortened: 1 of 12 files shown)


stephen-riggs (approved):
Looks fine to me. As discussed before, we need to be careful that the tests always do run on github, but I think the method you have here is fine.

I tried running the tests on my machine and found they crash if you have `murfey_dls_extras` installed, but otherwise all seemed good.

PR #664, opened 2025-09-09
Fixes for tomo multigrid destinations

stephen-riggs:
- Destination should be from sample not d
- Name tomo metadata as `metadata_Supervisor_..._SampleX`
- Match fractions and metadata in the destinations directory


d-j-hatton (approved):

d-j-hatton:
Thanks

PR #682, opened 2025-10-15
Feedback from tomogram picking

stephen-riggs:
Ability to receive messages back from tomogram picking and then send on a request to run tomogram 2d extraction/classification.

- Registers new pj recipe for the first tomogram
- Changes `SPAFeedbackParameters` to a more general `ClassificationFeedbackParameters` for reuse
- Add table for tomogram picks

Currently then set to run 2D classification in batches of 10000 particles with 5 classes. Needs deciding if that is the right value.
Also probably isn't sending exactly the right set of parameters onwards to the next recipe as the recipe hasn't yet been written, but the right items should be stored in the database here.

### DIFF
--- a/src/murfey/util/processing_params.py
+++ b/src/murfey/util/processing_params.py
@@ -82,7 +82,7 @@ default_spa_parameters = SPAParameters()
 
 
 class TomographyParameters(BaseModel):
-    batch_size_2d: int = 10000
+    batch_size_2d: int = 5000
     nr_classes_2d: int = 5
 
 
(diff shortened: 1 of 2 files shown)


src/murfey/workflows/tomo/picking.py
```diff
@@ -0,0 +1,208 @@
...
+    Session as MurfeySession,
+    TomogramPicks,
+    TomographyProcessingParameters,
+)
+from murfey.util.processing_params import default_tomo_parameters
+
+logger = getLogger("murfey.workflows.tomo.feedback")
+
+
+def _ids_tomo_classification(app_id: int, recipe: str, _db) -> Tuple[int, int]:
```
tieneupin: For database parameters like this, we previously settled on using `murfey_db` to identify which database we're calling. Perhaps we could continue with that convention here as well?

PR #683, opened 2025-10-16
Raise an exception if no 2D murfey ids present

stephen-riggs:
We keep hitting a case where no ids get sent to 2D classification. It's not clear why this happens, but if we don't have ids to send it would be better to just raise an exception in murfey.

Also I've changed the fixed number of 50 ids to the value in the default parameters (which is also 50 currently).

### DIFF
--- a/src/murfey/server/feedback.py
+++ b/src/murfey/server/feedback.py
@@ -258,6 +258,8 @@ def _3d_class_murfey_ids(particles_file: str, app_id: int, _db) -> Dict[str, int
             and db.Class3D.pj_id == pj_id
         )
     ).all()
+    if not classes:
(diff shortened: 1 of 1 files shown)


src/murfey/server/feedback.py
```diff
@@ -241,6 +241,8 @@ def _2d_class_murfey_ids(particles_file: str, app_id: int, _db) -> Dict[str, int
             db.Class2D.particles_file == particles_file and db.Class2D.pj_id == pj_id
         )
     ).all()
+    if not classes:
+        raise ValueError(f"No 2D classification IDs found for {particles_file}")
```
tieneupin: Would it be worth applying the same check to the `_3d_class_murfey_ids()` function immediately below this one?

PR #689, opened 2025-10-31
Register CLEM data in ISPyB

tieneupin:
Adds logic to register incoming CLEM data to both ISPyB and its mirror tables on Murfey.
* Adds new columns to the `CLEMImageSeries` table to keep track of image metadata needed for the ISPyB insert
* Adds logic to create a data collection group entry for CLEM data in ISPyB
* Adds logic to register grid overview datasets as atlases in ISPyB
* Adds logic to register positional datasets as grid squares in ISPyB
* Migrates hard-coded image scaling logic in some of the ISPyB insertion functions into the workflows that scale them as such, allowing the insertion function to be more universally used

### DIFF
--- a/src/murfey/workflows/clem/register_preprocessing_results.py
+++ b/src/murfey/workflows/clem/register_preprocessing_results.py
@@ -501,13 +501,12 @@ def run(message: dict, murfey_db: Session, demo: bool = False) -> dict[str, bool
             )
 
         try:
-            # Register dataset as grid square
-            if "Overview_" not in result.series_name:
-                _register_grid_square(
-                    session_id=session_id,
-                    result=result,
-                    murfey_db=murfey_db,
-                )
+            # Register CLEM image series as grid squares
+            _register_grid_square(
+                session_id=session_id,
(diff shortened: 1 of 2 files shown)


src/murfey/util/db.py
```diff
@@ -270,18 +287,27 @@ class CLEMImageSeries(SQLModel, table=True):  # type: ignore
...
 
-    # Process checklist for series
-    number_of_members: int = (
-        0  # Expected number of image stacks belonging to this series
-    )
+    # Shape and resolution information
+    pixels_x: Optional[int] = Field(default=None)
+    pixels_y: Optional[int] = Field(default=None)
+    pixel_size: Optional[float] = Field(default=None)
+    units: Optional[str] = Field(default=None)
```
stephen-riggs: Do you need the units in the database? I can't see they are used anywhere

tieneupin: That's a valid point. I included it as a just-in-case, so I'll revisit this in the future if no new workflows end up using this column.

PR #695, opened 2025-11-20
Register dcg for atlases

stephen-riggs:
Add data collection groups into ispyb for atlases. Each time the client sees a new atlas xml then a dcg is made.

Needs to be merged with #696


tieneupin (commented):
Thanks for doing this. It looks good to me, and works with the expected behaviour.

tieneupin:
We can probably close this since it was merged into PR #696.

PR #696, opened 2025-11-21
Match data collection groups for atlases and processing

stephen-riggs:
Needs to merge with #695 

Creates a common function for data collection group creation by the client. The data collection groups are made as atlases then updated to be SPA or Tomo if data collection happens.

Diff looks bigger than it really is as a test file got moved.


src/murfey/client/multigrid_control.py
```diff
@@ -12,6 +12,7 @@
 
 import murfey.client.websocket
 from murfey.client.analyser import Analyser
+from murfey.client.context import ensure_dcg_exists
 from murfey.client.contexts.spa import SPAModularContext
```
tieneupin: We shouldn't do it in this PR, but I reckon it's about time we change `SPAModularContext` to `SPAContext`?

PR #780, opened 2026-04-08
Register all grid squares from atlases

stephen-riggs:
Currently we only register EPU grid squares when data collection is started. This sets up a method to register them in all atlases as during screening.

The problem comes on conversion of screening sessions to EPU sessions. Here this loops through all grid squares and changes the tag when the data collection experiment type id is changed.

### DIFF
--- a/src/murfey/server/api/session_control.py
+++ b/src/murfey/server/api/session_control.py
@@ -359,7 +359,7 @@ def get_foil_hole(
 
 class AtlasRegistration(BaseModel):
     name: str
-    acquisition_uuid: str
+    acquisition_uuid: str | None
     storage_folder: str = ""
     register_grid: bool = False
     tag: str = ""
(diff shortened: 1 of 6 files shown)


src/murfey/server/api/session_control.py
```diff
@@ -371,7 +371,7 @@ def register_atlas(
     atlas_registration_data: AtlasRegistration,
     db=murfey_db,
 ):
-    if SMARTEM_ACTIVE:
+    if SMARTEM_ACTIVE and atlas_registration_data.acquisition_uuid is not None:
```
tieneupin, line 374: When would `acquisition_uuid` be None in normal operation?

tieneupin (approved):
Looks good! Just a question about when the Atlas' `acquisition_uuid` might end up as `None`.

src/murfey/server/api/session_control.py
stephen-riggs: Most of the time! `acquisition_uuid` will only be set if smartem is active

PR #787, opened 2026-04-20
Optimised database registration logic in CLEM workflow

tieneupin:
Fixes a couple of issues with the database registration logic in the CLEM workflow:

**Handle the possibility of there being multiple atlas images associated with a grid**
We occasionally run into situations where more than one image set for a grid meets the criteria to be registered as an atlas, which breaks the workflow because the SQL query expects a single matching database entry or no result. 

This can be resolved by returning  _all_ matching atlas entries, sorted by insertion order. The latest one is used to perform the ISPyB updates. Because the `GridSquare` and `DataCollectionGroup` entries are re-updated for every new image set received, all entries for a given sample grid will be re-updated correctly with reference to the last atlas registered.

**Accidental insert when updating `ImagingSite` entries after registering `GridSquare` entries**
There was a race condition between when the `ImagingSite` entries from `_register_clem_imaging_site` are committed and when they are queried in `_register_dcg_and_atlas`. This lead to the creation of duplicate `ImagingSite` entries in the database. This can be fixed by modifying the subsequent helper functions to use the registered `ImagingSite` entry from `_register_clem_imaging_site` instead of querying it from the database.

**Less aggressive commits**
In `register_grid_square`, `db.commit()` was previously run multiple times within a single `for` loop. By moving it out of the loop, this should improve performance.

### DIFF
--- a/src/murfey/workflows/clem/register_preprocessing_results.py
+++ b/src/murfey/workflows/clem/register_preprocessing_results.py
@@ -118,12 +118,20 @@ def _register_clem_imaging_site(
     result: CLEMPreprocessingResult,
     murfey_db: Session,
 ):
-    def _register(
+    """
+    Creates an ImagingSite database entry for the current CLEM preprocessing result
+    if one doesn't already exist, or modifies the existing one if it does. Each entry
+    corresponds to a unique site on the sample grid, and results containing denoised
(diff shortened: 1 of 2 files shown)


stephen-riggs (approved):
This looks cleaner than before, thanks for making the discussed changes

PR #790, opened 2026-04-27
Doppio live processing requirements

stephen-riggs:


Koldstream:
I've had a good look over and I feel pretty happy with it! I had forgotten how many changes had been made overall.

As discussed on Friday the logic for the paths in processing_params.py will need looking at, and further testing for the job reservation system... I'll have another look at this soon.

src/murfey/util/processing_params.py
```diff
@@ -1,36 +1,102 @@
...
-        / "MotionCorr"
-        / "job002"
-        / "Movies"
-        / ppath.parent.relative_to(core / sub_dataset)
-        / str(ppath.stem + "_motion_corrected.mrc")
-    )
+        parts = [secure_filename(p) for p in input_movie.parts]
+        visit_idx = parts.index(visit_name)
+        core = Path("/") / Path(*parts[: visit_idx + 1])
+        ppath = Path("/") / Path(*parts)
```
tieneupin: From the limited testing I've done, it looks like `Path(*parts...)` will give you the same result as `Path("/") / Path(*parts...)`, so this could actually be simplified.

src/murfey/server/api/workflow.py
```diff
@@ -504,8 +509,16 @@ def register_spa_proc_params(
...
+    if murfey.server._transport_object:
+        murfey.server._transport_object.send(
+            murfey.server._transport_object.feedback_queue, zocalo_message
+        )
+    else:
+        logger.error(
+            f"Pre-processing was requested for {sanitise(str(session_id))} "
+            "but no Zocalo transport object was found"
+        )
+        return proc_params
```
tieneupin: If I'm reading this correctly, it will return `None` when successful and `proc_params` if unsucessful. Is this intended behaviour?

src/murfey/server/api/workflow.py
```diff
@@ -119,7 +119,7 @@ def register_dc_group(
     visit_name: str,
     session_id: MurfeySessionID,
     dcg_params: DCGroupParameters,
-    db=murfey_db,
+    db: Session = murfey_db,
```
tieneupin: You appear to be using `murfey.util.db.Session` as the type hint, which is an SQLModel table. I'm not sure how the tests have passed, but this should be pointing to `sqlmodel.Session` instead.

This would be where I'd usually alias the SQLModel Session and Murfey DB Session so that they are clearly distinguishable.

e.g.
```
from sqlmodel import Session as SQLModelSession

import murfey.util.db as MurfeyDB
```

src/murfey/server/api/session_info.py
```diff
@@ -131,7 +131,7 @@ def all_visit_info(
 
 
 @router.get("/sessions/{session_id}/rsyncers", response_model=List[RsyncInstance])
-def get_rsyncers_for_client(session_id: MurfeySessionID, db=murfey_db):
+def get_rsyncers_for_client(session_id: MurfeySessionID, db: Session = murfey_db):
```
tieneupin: Same error in this module. `Session` is referring to `murfey.util.db.Session`, which is an SQLModel table. The type hint should be pointing to `sqlmodel.Session`.

src/murfey/server/api/workflow.py
```diff
@@ -568,13 +583,15 @@ async def request_spa_preprocessing(
...
-        feedback_params = params[1]
+        # SPARelionParameters is an ORM row, but the recipe parameters below are
+        # read by key, so convert to a dict.
+        proc_params: dict | None = dict(
+            db.exec(
+                select(SPARelionParameters).where(
+                    SPARelionParameters.pj_id == collected_ids[2].id
+                )
+            ).one()
+        )
```
tieneupin: This is an SQLModel table, so it's apparently possible to use `table.model_dump()` to convert this into a dictionary structure.

```
proc_params_query = db.exec(
    select(SPARelionParameters).where(
        SPARelionParameters.pj_id == collected_ids[2].id
    )
).one_or_none()
proc_params = proc_params_query.model_dump() if proc_params_query else None
```

EDIT: If `proc_params` attributes can be called directly later on in this function, the `dict` conversion here will become unnecessary.

src/murfey/server/api/workflow.py
tieneupin: In this endpoint function, `proc_params` _doesn't_ have to be read in as a `dict` does it? This would also work with `"kv": proc_params.voltage`, I would imagine.

tieneupin: Same here.
`"pixel_size": proc_params.angpix`

tieneupin:
```
                "ft_bin": proc_params.motion_corr_binning,
                "fm_dose": proc_params.dose_per_frame,
                "gain_ref": proc_params.gain_ref,
```

tieneupin: `"particle_diameter": proc_paramsparticle_diameter or 0`

tieneupin, line 727:
```
                "fm_int_file": (
                    proc_params.eer_fractionation_file
                    if proc_params.eer_fractionation_file
                    else proc_file.eer_fractionation_file
                ),
```

tests/workflows/fib/test_register_atlas.py
```diff
@@ -363,12 +363,10 @@ def test_run_with_db(
...
     mocker.patch(
-        "murfey.workflows.register_data_collection_group._transport_object",
-        new=TransportManager("PikaTransport"),
+        "murfey.server._transport_object", new=TransportManager("PikaTransport")
     )
     mocker.patch(
-        "murfey.workflows.register_atlas_update._transport_object",
-        new=TransportManager("PikaTransport"),
+        "murfey.server._transport_object", new=TransportManager("PikaTransport")
     )
```
tieneupin: Now that we're patching `murfey.server._transport_object` directly, only one `mocker.patch` call is needed.

src/murfey/util/processing_params.py
stephen-riggs: ok, done

src/murfey/server/api/workflow.py
stephen-riggs: Not really intended so will standardise, although we don't use the return value

src/murfey/server/api/workflow.py
stephen-riggs: Done

src/murfey/server/api/session_info.py
stephen-riggs: Done

src/murfey/server/api/workflow.py
stephen-riggs: Agreed, no reason to dictionary conver it

tests/workflows/fib/test_register_atlas.py
stephen-riggs: done

tieneupin (approved):
Looks good to me now. Nice work! 👍

PR #795, opened 2026-04-30
Add sxt router and rename recipe

stephen-riggs:
Bugfixes for the sxt setup which have appeared in testing


tieneupin (approved):

tieneupin:
It's worth noting that the recipe names are changed because we need to identify the software being used for the tomo alignment.

PR #796, opened 2026-04-30
Update FIBContext and backend logic for detecting, handling and making GIFs of drift correction images

tieneupin:
This PR updates the logic on the client and backend server used to construct GIFs of the drift correction images.
* Creates and stores the output GIF file path once per site on the client side instead of reconstructing it each time on the server side
* Uses the site info metadata parsed from the 'ProjectData.dat' file to determine the output GIF file path
* Removes timestamp extraction logic, as the AutoTEM's drift correction images are automatically generated with the timestamp in the file name in `yyyy-mm-dd-hh-mm-ss` convention, which allow them to be sorted as strings
* Adds logic to skip sending GIF creation requests if the metadata is not yet parsed, sending it only after the corresponding metadata information has been collected
* Simplifies the API endpoint URL path and the parameters that need to be passed to the backend

Diff due to tests is about +270/-110.


src/murfey/client/contexts/fib.py
```diff
@@ -289,82 +288,75 @@ def post_transfer(
...
-                    session_id=environment.murfey_session,
-                )
+                    # Post drift correction GIF request if it hasn't already been done
+                    if (
+                        fib_image := self._drift_correction_images.get(site_num, None)
+                    ) is not None and not fib_image.is_submitted:
+                        # Construct the output file name if it doesn't already exist
+                        if (output_file := fib_image.output_file) is None:
+                            if not (
+                                source := _get_source(transferred_file, environment)
```
d-j-hatton: my preference here would be to have `source = _get_source(transferred_file, environment)` on the line above rather than using the walrus. at the moment `source` gets used outside of this branch and it feels weird having the assignment fall through in a kind of uncontrolled manner

src/murfey/client/contexts/fib.py
```diff
@@ -289,82 +288,75 @@ def post_transfer(
...
+                                        self._machine_config.get("rsync_basepath", "")
+                                    ),
+                                )
+                            ):
+                                logger.warning(
+                                    f"Could not find destination file path for {transferred_file.name!r}"
+                                )
+                                continue
+                            if (
+                                output_dir := self._determine_output_dir(
```
d-j-hatton: same as above

src/murfey/server/api/workflow_fib.py
```diff
@@ -1,14 +1,20 @@
 import json
 import logging
+import os
 from importlib.metadata import entry_points
 from pathlib import Path
 
+import numpy as np
+import PIL.Image
```
d-j-hatton, line 8: are we guaranteed to have `pillow` now? (just asking because in the old implementation there was an import guard)

tieneupin: Yeah, Pillow is guaranteed to be present on the backend. We've included it as a mandatory dependency in `pyproject.toml`.

d-j-hatton: cool

src/murfey/client/contexts/fib.py
tieneupin: That's fair. I've rewritten the sections where walrus equalities were used counter-intuitively.

d-j-hatton: thanks

d-j-hatton (approved):

PR #798, opened 2026-05-12
Add endpoints to get, create and delete silences

Abigail-Yates:
Turn alerts off and on using silences


src/murfey/server/api/session_info.py
```diff
@@ -276,6 +277,65 @@ def update_current_gain_ref(
...
+    machine_config = machine_info_by_instrument(instrument_name)
+    alertmanager_url = machine_config.alertmanager_url
+    start_time = datetime.now().astimezone().isoformat()
+    end_time_str = end_time.astimezone().isoformat()
+    silence_json = {
+        "matchers": [
+            {"name": "microscope", "value": instrument_name, "isRegex": False}
+        ],
+        "createdBy": "murfey",
+        "annotations": {"description": "Test"},
```
stephen-riggs: Description could be something like "Silence for alerts on {instrument_name}" ?

src/murfey/server/api/session_info.py
```diff
@@ -276,6 +277,65 @@ def update_current_gain_ref(
...
+)  # delete all silences for given microscope
+def delete_silences(instrument_name: MurfeyInstrumentName):
+    machine_config = machine_info_by_instrument(instrument_name)
+    alertmanager_url = machine_config.alertmanager_url
+    silences = get_silences(instrument_name)
+    if len(silences) == 0:
+        return None
+    for silence in silences:
+        id = silence["id"]
+        response = requests.delete(f"{alertmanager_url}/api/v2/silence/{id}")
```
stephen-riggs, line 360: Possibly this should log something if the response status code isn't 200, so we would be able to see that?

src/murfey/util/config.py
```diff
@@ -118,6 +118,9 @@ class MachineConfig(BaseModel):  # type: ignore
     # Pydantic BaseModel settings
     model_config = ConfigDict(extra="allow")
 
+    # Alertmanager URL
+    alertmanager_url: str = "https://murfey-alertmanager.diamond.ac.uk"
```
stephen-riggs: This might be better under the "Configuration and URLs" section of the config. And the default shouldn't be the diamond address

src/murfey/server/api/session_info.py
```diff
@@ -276,6 +277,87 @@ def update_current_gain_ref(
...
+    all_configs = get_machine_config()
+    valid_instrument_names = list(all_configs.keys())
+    if instrument_name in valid_instrument_names:
+        index = valid_instrument_names.index(instrument_name)
+        query_params = f"filter=microscope={sanitise(valid_instrument_names[index])}"
+    else:
+        return None
+    machine_config = all_configs[instrument_name]
+    alertmanager_url = machine_config.alertmanager_url
+    if alertmanager_url == "":
```
stephen-riggs: It's more standard in python to do this check using `if not alertmanager_url:` as that would check for all "Falsey" cases like `None`, `[]` or `""`

src/murfey/server/api/session_info.py
```diff
@@ -276,6 +277,87 @@ def update_current_gain_ref(
...
+
+@router.get("/silences/{instrument_name}")
+def get_silences(instrument_name: MurfeyInstrumentName):
+    all_configs = get_machine_config()
+    valid_instrument_names = list(all_configs.keys())
+    if instrument_name in valid_instrument_names:
+        index = valid_instrument_names.index(instrument_name)
+        query_params = f"filter=microscope={sanitise(valid_instrument_names[index])}"
+    else:
+        return None
```
stephen-riggs: Fine like this, but if the function return is list or None the frontend would need to be able to handle either

stephen-riggs (approved):
Looks good to me now, thanks for addressing all the comments

PR #801, opened 2026-05-15
SXT bugfixes and improvements

stephen-riggs:
Various changes needed to make the SXT workflow smoother:
- Use olefile directly and remove the txrm2tiff dependency
- Read energy, tilt limits and tilt count from the txrm metadata and store them in the `DataCollection` ispyb table
- Skip processing of reference txrm files
- Allow processing of nested folders
- Set session file to None and then tag the DCG with the unaltered metadata path
- Set the data collection comment as any part of the file name before `_angle`

The remaining issue is how to set the rsyncers correctly, which we need to discuss


src/murfey/client/contexts/sxt.py
```diff
@@ -66,12 +61,18 @@ def register_sxt_data_collection(
                 "source": str(self._basepath),
                 "tag": tilt_series,
                 "pixel_size_on_image": str(
-                    data_collection_parameters.get("pixel_size", 100)
+                    round(data_collection_parameters.get("pixel_size", 100), 2) * 1e-10
```
tieneupin, line 64: Why does it need to be kept to 2 decimal places?

src/murfey/client/contexts/sxt.py
```diff
@@ -46,12 +44,9 @@ def register_sxt_data_collection(
             )
             return
         try:
-            metadata_source = (
-                self._basepath.parent / environment.visit / self._basepath.name
-            )
             ensure_dcg_exists(
                 collection_type="sxt",
-                metadata_source=metadata_source,
+                metadata_source=self._basepath,
```
tieneupin, line 49: Is this related to "[allowing] the processing of nested folders"?

src/murfey/client/contexts/sxt.py
```diff
@@ -127,84 +128,103 @@ def post_transfer(
...
-                    "ImageInfo/PixelSize",
-                    XrmDataTypes.XRM_FLOAT,
-                    strict=True,
-                )
-                if pixel_size_txrm:
+                if txrm_ole.exists("ImageInfo/PixelSize"):
+                    pixel_size_txrm = np.frombuffer(
+                        txrm_ole.openstream("ImageInfo/PixelSize").getvalue(),
+                        np.float32,
+                    ).tolist()
```
tieneupin, line 150: Will these values be lists of floats/ints at this point? If yes, will there be more than one value in the list, and will they be distinct?

src/murfey/client/contexts/sxt.py
```diff
@@ -127,84 +128,103 @@ def post_transfer(
...
+                        metadata["energy"] = int(round(axis_values[energy_index]))
+
+            if not metadata.get("has_reference", False):
+                logger.debug(f"Reference image {transferred_file} not processed")
+                return True
 
+            visit_index = transferred_file.parent.parts.index(environment.visit)
+            destination_search_dir = "/".join(
+                transferred_file.parent.parts[: visit_index + 2]
+            ).replace("//", "/")
```
tieneupin, line 216: `.parent` is not strictly needed, as the indexing and slicing should work the same on the original `transferred_file` variable.

src/murfey/client/contexts/sxt.py
```diff
@@ -227,11 +247,13 @@ def post_transfer(
                 visit_name=environment.visit,
                 session_id=environment.murfey_session,
                 data={
-                    "session_id": environment.murfey_session,
                     "tag": transferred_file.stem,
-                    "source": str(transferred_file.parent),
-                    "pixel_size": metadata.get("pixel_size", 100),
+                    "source": destination_search_dir,
+                    "pixel_size": round(metadata.get("pixel_size", 100), 2),
```
tieneupin: Probably worth adding a comment here to note what units pixel size will be. Based on the `100`, are we expecting it to be consistently in nanometres?

tieneupin (approved):
Some requests for clarification, but beyond that, the changes looks good. Nice work!

src/murfey/client/contexts/sxt.py
stephen-riggs: It doesn't need to be, it just seemed cleaner to do so as it avoids python float issues

src/murfey/client/contexts/sxt.py
stephen-riggs: The txrm return is a list of one value

src/murfey/client/contexts/sxt.py
stephen-riggs: Pixel size is angstroms, I can comment that

src/murfey/client/contexts/sxt.py
stephen-riggs: Yes, and the previous behaviour was also unnecessary in this case. It was previously trying to do the path construction needed on the Krios instruments

src/murfey/client/contexts/sxt.py
tieneupin: Is there a need to keep it as a list of one, then?

stephen-riggs: It's not being kept as a list - the following line takes the value and puts it into the metadata dictionary as a single float

tieneupin: Oops, that's my bad. Thanks for clarifying.

src/murfey/client/contexts/sxt.py
```diff
@@ -127,84 +128,103 @@ def post_transfer(
...
-                    XrmDataTypes.XRM_FLOAT,
-                    strict=True,
-                )
-                if pixel_size_txrm:
+                if txrm_ole.exists("ImageInfo/PixelSize"):
+                    pixel_size_txrm = np.frombuffer(
+                        txrm_ole.openstream("ImageInfo/PixelSize").getvalue(),
+                        np.float32,
+                    ).tolist()
                     metadata["pixel_size"] = pixel_size_txrm[0] * 1e4
```
tieneupin, line 151: Are you converting to micrometres here?

stephen-riggs: Looks like I set it to go from microns to angstroms. That was done in a previous PR but I guess the txrm is storing it in microns then

PR #803, opened 2026-05-19
Add a config key to allow transfer to a single raw directory

stephen-riggs:
One possible option for how to resolve the b24 data transfer where we want everything to go to the same raw folder, but using different rsyncers and analysers.

This adds a config key to override the suggested path logic and return raw each time. It also turns the limited transfer behaviour off in for all rsyncers made by this method so the files are analysed.

### DIFF
--- a/src/murfey/client/watchdir_multigrid.py
+++ b/src/murfey/client/watchdir_multigrid.py
@@ -49,12 +49,12 @@ class MultigridDirWatcher(Observer):
             self.thread.join()
         log.debug("MultigridDirWatcher thread stop completed")
 
-    def _handle_metadata(self, directory: Path, extra_directory: str):
+    def _handle_metadata(self, directory: Path, extra_directory: str, limited=True):
         self.notify(
             directory,
             extra_directory=extra_directory,
             analyse=self._analyse,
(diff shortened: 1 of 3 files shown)


src/murfey/client/watchdir_multigrid.py
```diff
@@ -49,12 +49,12 @@ def stop(self):
             self.thread.join()
         log.debug("MultigridDirWatcher thread stop completed")
 
-    def _handle_metadata(self, directory: Path, extra_directory: str):
+    def _handle_metadata(self, directory: Path, extra_directory: str, limited=True):
```
tieneupin, line 52: I think `_handle_subdir` might be a suitable replacement name for `_handle_metadata`, as it seems to be notifying the notifying the subscribed components to create an additional Analyser for the sub-directory stated in `extra_directory`.

tieneupin (approved):
Code seems fine!

src/murfey/client/watchdir_multigrid.py
stephen-riggs: Not exactly, as discussed the name is unchanged but a comment has been added

PR #805, opened 2026-05-20
Register FIB data in ISPyB

tieneupin:
Extends the existing FIB data registration workflow to insert a DataCollectionGroup and Atlas into ISPyB for each unique sample detected.

Currently inserts the path to the real image, as thumbnail generation logic hasn't been implemented yet.


src/murfey/workflows/fib/register_atlas.py
```diff
@@ -150,36 +153,147 @@ def _register_fib_imaging_site(
...
+            "dcgid": dcg_entry.id,
+            "atlas_id": dcg_entry.atlas_id,
+            "atlas": atlas_name,
+            "atlas_pixel_size": atlas_pixel_size,
+            "sample": dcg_entry.sample,
+        }
+        if entry_point_result := entry_points(
+            group="murfey.workflows", name="atlas_update"
+        ):
+            (workflow,) = entry_point_result
```
stephen-riggs, line 297: Does this (and the dcg one below) need to be done here rather than sending a message to the feedback queue?

src/murfey/workflows/fib/register_atlas.py
```diff
@@ -150,36 +153,147 @@ def _register_fib_imaging_site(
...
+    metadata: FIBAtlasMetadata,
+    murfey_db: Session,
+):
+    proposal_code = "".join(char for char in visit_name.split("-")[0] if char.isalpha())
+    proposal_number = "".join(
+        char for char in visit_name.split("-")[0] if char.isdigit()
+    )
+    visit_number = visit_name.split("-")[-1]
+
+    # Register using thumbnail values if they are provided
```
stephen-riggs, line 269: What circumstance do you want thumbnail vs full image sizes? Could these ever get mixed up?

src/murfey/workflows/fib/register_atlas.py
```diff
@@ -150,36 +153,147 @@ def _register_fib_imaging_site(
...
+            )
+        else:
+            logger.warning("No workflow found for 'data_collection_group'")
+    dcg_entry = murfey_db.exec(
+        select(MurfeyDB.DataCollectionGroup)
+        .where(MurfeyDB.DataCollectionGroup.session_id == session_id)
+        .where(MurfeyDB.DataCollectionGroup.tag == imaging_site.site_name)
+    ).one()
+
+    imaging_site.dcg_id = dcg_entry.id
```
stephen-riggs, line 334: Looks like you do need the dcg workflow run, this id is a bit challenging

src/murfey/workflows/fib/register_atlas.py
```diff
@@ -219,6 +335,26 @@ def run(
...
+            _register_dcg_and_atlas(
+                session_id=session_id,
+                instrument_name=murfey_session.instrument_name,
+                visit_name=murfey_session.visit,
+                imaging_site=fib_imaging_site,
+                metadata=metadata,
+                murfey_db=murfey_db,
+            )
+        except Exception:
+            # Log error but allow workflow to proceed
```
stephen-riggs, line 421: Do you want this to proceed? If you fail would it be better to dlq the message?

stephen-riggs (approved):
This all looks like it will work, just a few queries about the structure

src/murfey/workflows/fib/register_atlas.py
tieneupin: As discussed, we'll DLQ the message instead.

src/murfey/workflows/fib/register_atlas.py
tieneupin: As discussed, fully modularising this is too much effort for too little gain at the moment. However, I have rewritten this workflow so that it is triggered by the `feedback_callback` function instead, and can be safely acked or nacked.

src/murfey/workflows/fib/register_atlas.py
tieneupin: The intent is to ensure that the image and pixel size registered in ISPyB are consistent with one another. As long as that happens, image overlaying and alignment logic should work regardless of whether the original or thumbnail image was registered and used for the job.

src/murfey/workflows/fib/register_atlas.py
```diff
@@ -142,6 +157,35 @@ def _parse_metadata(file: Path, visit_name: str):
...
+    img.thumbnail((512, 512))
+
+    # Find visit directory path
+    visit_idx = file.parts.index(visit_name)
+    visit_dir = list(reversed(file.parents))[visit_idx]
+
+    # Construct processed directory and set permissions
+    processed_dir = visit_dir / "processed"
+    processed_dir.mkdir(exist_ok=True)
+    os.chmod(processed_dir, mode=mode)
```
stephen-riggs: I think we need to discuss whether the chmod is the right thing to do here

tieneupin: As discussed, will remove this logic and let the processed directory inherit permissions normally for now

stephen-riggs (approved):
Thanks for addressing my comments

PR #806, opened 2026-05-21
add and set alert_end_time

Abigail-Yates:
- Add alert_end_time as a prometheus metric. It will be a unix timestamp. 
- Set alert_end_time when session is created (if one exists and is further in the future than the current alert_end_time)
- Update alert_end_time when session end time is updated 

(also updated one log warning statement in the create silence endpoint to make it consistent with the others)

### DIFF
--- a/src/murfey/server/api/instrument.py
+++ b/src/murfey/server/api/instrument.py
@@ -246,6 +246,13 @@ async def update_visit_end_time(
     db.add(session_entry)
     db.commit()
 
+    # Update the alert endtime on prometheus
+    alert_end_time = prom.alert_end_time._value.get()
+    if end_time:
+        visit_end_timestamp = end_time.timestamp()
+        if alert_end_time < visit_end_timestamp:
+            prom.alert_end_time.set(visit_end_timestamp)
+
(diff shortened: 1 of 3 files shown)


src/murfey/server/api/instrument.py
```diff
@@ -246,6 +246,13 @@ async def update_visit_end_time(
     db.add(session_entry)
     db.commit()
 
+    # Update the alert endtime on prometheus
+    alert_end_time = prom.alert_end_time._value.get()
+    if end_time:
+        visit_end_timestamp = end_time.timestamp()
+        if alert_end_time < visit_end_timestamp:
```
stephen-riggs, line 253: Do you know what the alert_end_time value gives if it has never been set previously? Does it come up as 0 or is it None/NAN/undefined?

Abigail-Yates: On the /metrics page when it first runs it shows as 0.0

stephen-riggs: Ok, the < should be fine then

stephen-riggs (approved):
All looks good to me

PR #809, opened 2026-05-28
set alert_end_time for each visit separately

Abigail-Yates:
- added "visit" label to alert_end_time so each visit now has a different alert_end_time
    -> therefore don't need to check if new alert_end_time is longer than existing time (when creating or updating visit end time)
- remove alert_end_time for that visit when session removed


stephen-riggs (approved):
Thanks, this looks good to me

PR #811, opened 2026-05-29
Turn all nacks to requeue False

stephen-riggs:
A temporary workaround for the rabbitmq errors we are seeing


tieneupin (approved):
Looks good. Thanks for doing this.

PR #812, opened 2026-06-02
Basic SerialEM grid square mag image analyser

d-j-hatton:
Add basic functionality to trigger SmartEM events when grid square mag images are seen from a SerialEM workflow (assumes that the stem of the file name is the UUID associated with the grid square)


stephen-riggs (approved):
Looks fine to me. Are you intending to save the files into an Images-Disc directory like EPU does then?

d-j-hatton:
> Looks fine to me. Are you intending to save the files into an Images-Disc directory like EPU does then?

oops, thanks for the catch, no I'm not. fixed now

PR #815, opened 2026-06-03
Register FIB milling progress

tieneupin:
Adds new database tables + relationships and workflow functions to support the registration and tracking of the FIB milling process. The AutoTEM metadata read by Murfey will now be registered in ISPyB's new `MillingStep` table, and a local Murfey copy will be created so that Murfey can propagate updates to existing rows.

As part of this PR, the ISPyB schema and Python package versions have been bumped to make use of the newly introduced `MillingStep` table.

The diff looks huge, but +540 of those lines are due to tests.


src/murfey/workflows/fib/register_milling_progress.py
```diff
@@ -0,0 +1,454 @@
...
+
+    try:
+        # Parse and unpack incoming message
+        session_id = int(message["session_id"])
+        site_info = LamellaSiteInfo(**message["site_info"])
+        logger.debug(
+            "Received the following FIB metadata for registration:\n"
+            f"{json.dumps(site_info.model_dump(exclude_none=True), indent=2, default=str)}"
+        )
+    except Exception:
```
stephen-riggs, line 358: There's several bare Except clauses in the run function which would be better excepting specific things

stephen-riggs (approved):
Too big to review thoroughly but the general workflow looks fine

PR #819, opened 2026-06-11
Match dcgs for tomo multigrid

stephen-riggs:
The data collection groups for atlas, metadata and fractions were not being matched up correctly for tomo multigrid.

This is due to the metadata being in `Supervisor/Sample` and the data in `Supervisor_Sample`

This resolves this issue by always using the metadata form with a slash. In the tomo context it checks the existence of the different metadata locations and determines the `metadata_source` and `group_tag` from this.

Also catches errors from being unable to read grid squares on a tomo atlas, and not copying the batch positions list.

### DIFF
--- a/src/murfey/client/contexts/tomo.py
+++ b/src/murfey/client/contexts/tomo.py
@@ -109,6 +109,14 @@ class TomographyContext(Context):
             metadata_source = (
                 self._basepath.parent / environment.visit / self._basepath.name
             )
+            multigrid_metadata_source = (
+                self._basepath.parent
+                / environment.visit
+                / "_".join(self._basepath.name.split("_")[:-1])
+                / self._basepath.name.split("_")[-1]
+            )
+            if not metadata_source.exists() and multigrid_metadata_source.exists():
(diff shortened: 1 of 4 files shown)


src/murfey/client/contexts/tomo.py
```diff
@@ -109,6 +110,20 @@ def register_tomography_data_collections(
             metadata_source = (
                 self._basepath.parent / environment.visit / self._basepath.name
             )
+            multigrid_metadata_source = (
+                self._basepath.parent
+                / environment.visit
+                / "_".join(self._basepath.name.split("_")[:-1])
+                / self._basepath.name.split("_")[-1]
+            )
+            if not metadata_source.exists() and multigrid_metadata_source.exists():
```
d-j-hatton: should this be a more explicit check that `is_dir()`?

stephen-riggs: Yes that makes sense. Done

d-j-hatton (approved):

PR #821, opened 2026-06-12
Back-populate bfactor table into processing job

stephen-riggs:


tieneupin (approved):
That looks like it's implemented correctly.

PR #823, opened 2026-06-12
Support ICC computational clearing mode for CLEM datasets

tieneupin:
We were informed of the existence of another computational clearing (CC) mode offered by the CLEM.

This PR updates the workflow so that this CC mode is also supported. Slight improvements were made to the code to handle type hinting centrally via `TypeAlias` objects instead of manually typing out the identical, long types each time. 

Instead of looking for the presence of a string pattern associated with the CC mode in the series name, the logic was adjusted so that this string pattern is only valid if found at the _end_ of the series name.

Additionally, we switch from using the term 'denoising' to 'computational clearing'/'cc', as that is a more technically accurate indicator of the processing step's purpose. It does not subtract or remove noise, but rather identifies and subtracts out-of-focus background fluorescence from fluorescent images.

### DIFF
--- a/src/murfey/workflows/clem/register_preprocessing_results.py
+++ b/src/murfey/workflows/clem/register_preprocessing_results.py
@@ -33,7 +33,7 @@ ColorChannels: TypeAlias = Literal[
     "gray", "red", "green", "blue", "cyan", "magenta", "yellow"
 ]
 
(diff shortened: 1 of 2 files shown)


stephen-riggs (approved):
Looks good to me, I like the change to be more generic so that denoising modes and colours can be added in a single place if needed

PR #825, opened 2026-06-16
Update FIB metadata parsing and registration logic

tieneupin:
Previously, we forcibly mapped certain FIB steps to certain stage value fields. However, as Murfey's FIB workflow is run on more types of FIB datasets, we find that this hard mapping is incorrect in certain cases. Additionally, sites set up and selected outside of AutoTEM having a different naming pattern than if they were configured completely using AutoTEM.

This PR makes the metadata registration more flexible by using the _latest_ available stage position name to perform the ISPyB and Murfey database registrations. It also updates the number extraction function to handle these additional site naming patterns

### DIFF
--- a/src/murfey/util/fib.py
+++ b/src/murfey/util/fib.py
@@ -2,7 +2,7 @@
 General functinos specific to the FIB workflow
 """
 
-import re
+from pathlib import Path
 
 
 def number_from_name(name: str) -> int:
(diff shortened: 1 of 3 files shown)


stephen-riggs (approved):
Looks good to me. Does this resolve the issue which caused the current FIB dlq messages?

tieneupin:
> Does this resolve the issue which caused the current FIB dlq messages?

This will only partially address it. The 'waffle' method workflow has a very different data structure compared to the full AutoTEM workflow. It will necessitate additional logic (potentially on the FIBContext side) to fully separate the two workflows and how files from each workflow are handled.

PR #826, opened 2026-06-16
Search for a txrm reference is no internal one is present

stephen-riggs:
If a txrm does not have an internal reference, instead of dismissing it search for a reference file in xrm format. Takes the most recent xrm which is not a mosaic image.

Still skips processing for txrms with less than 20 frames as these will be reference stacks.

Also changes the tilt series tags to remove angle naming, and skips transfer of "New folder"

Pairs with https://github.com/DiamondLightSource/cryoem-services/pull/271

### DIFF
--- a/src/murfey/client/watchdir_multigrid.py
+++ b/src/murfey/client/watchdir_multigrid.py
@@ -93,7 +93,9 @@ class MultigridDirWatcher(Observer):
     def _process(self):
         while not self._stopping:
             for d in self._basepath.glob("*"):
-                if d.name in self._machine_config["create_directories"]:
+                if d.name == "New folder":
+                    self._seen_dirs.append(d)
+                elif d.name in self._machine_config["create_directories"]:
                     if d.is_dir() and d not in self._seen_dirs:
                         self.notify(
                             d,
--- a/src/murfey/client/watchdir_multigrid.py
+++ b/src/murfey/client/watchdir_multigrid.py
@@ -93,7 +93,7 @@ class MultigridDirWatcher(Observer):
     def _process(self):
         while not self._stopping:
             for d in self._basepath.glob("*"):
-                if d.name == "New folder":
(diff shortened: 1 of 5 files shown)


src/murfey/client/watchdir_multigrid.py
```diff
@@ -93,7 +93,9 @@ def _handle_fractions(self, directory: Path):
     def _process(self):
         while not self._stopping:
             for d in self._basepath.glob("*"):
-                if d.name in self._machine_config["create_directories"]:
+                if d.name == "New folder":
+                    self._seen_dirs.append(d)
```
tieneupin, line 97: What does putting "New folder" in `_seen_dirs` do?

I wonder if it's possible or worth considering putting this logic in the `substrings_blacklist` `MachineConfig` key, as that will also allow for this Murfey to skip 'New folder' instances.

stephen-riggs: The aim is to never register "New folder" for transfer. This has been a problem on almost all instruments due to the way windows creates folders, so I think we just want to exclude them in all cases rather than using the blacklist

tieneupin: OK, that's fair. In that case, would it be worth taking this a step further to also consider the file patterns for subsequent 'New folder' instances? Windows will create "New folder (2)", "New folder (3)", and so on after the first instance.

tieneupin (approved):
Some ongoing discussion how to handle "New folder", but looks good otherwise!

src/murfey/client/watchdir_multigrid.py
stephen-riggs: Can do, I don't see any reason you would want to transfer "New folder (2)" etc

tieneupin:
Nice. Thanks for addressing my comment. Feel free to merge when you're ready.

PR #827, opened 2026-06-16
Match grid squares on sample rather than tag if sample known

stephen-riggs:
There is a problem with registering grid squares from an atlas, as the dcg tag is being used as the matching but this can change if atlas converts to SPA.

Instead, when registering squares from an atlas file, provide the sample and match this to a dcg. And send dcg request before that to make sure one will exist.


tieneupin (approved):
Looks good to me!

PR #258, opened 2024-04-23
Change the tomography alignment request methods to avoid cases of duplication

stephen-riggs:
This implements a check that processing has not already been requested, and also checks the length of the tilt list after motion correction.

Closes #257

PR #830, opened 2026-06-17
Allow multiple reconstruction recipes for sxt

stephen-riggs:
This moves the SXT processing into a rabbitmq message, and adds support for more recipes.
Recipes in the machine config are detected and all of these are used to send processing if the collection type is SXT. This will allow for using imod and aretomo.

Links to https://github.com/DiamondLightSource/cryoem-services/pull/258


tieneupin (approved):
Looks good to me!

PR #832, opened 2026-06-18
Fully overwrite values in MillingStep entry upon update

tieneupin:
Now that the FIB milling progress registration logic is live, we have been able to start detecting bugs that weren't caught during testing.

In the FIB AutoTEM metadata, once a step is complete, the `is_enabled` field is set back to `False`. However, the current ISPyB insertion logic only supports the insertion and updating of truth-y values. Once the field is set to `True`, it can't be set back to `False` again based on current logic.

This PR fixes that by fully overwriting all values in an existing `MillingStep` database entry when updating it, instead of updating the value only when it's truth-y.

### DIFF
--- a/src/murfey/server/ispyb.py
+++ b/src/murfey/server/ispyb.py
@@ -813,6 +813,10 @@ class TransportManager:
         width_overlap_rear_left: float | None = None,
         width_overlap_rear_right: float | None = None,
     ):
+        """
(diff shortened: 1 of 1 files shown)


tieneupin:
Discussed and not needed. It's more practical to have a record of steps that were previously enabled than it is to fully follow the metadata.

PR #833, opened 2026-06-22
Skip manual AutoTEM

tieneupin:
We have established that it is possible to use AutoTEM in manual mode and in full auto mode. The manual AutoTEM mode uses a different data structure and naming patterns, causing errors with our current file parsing logic, and will need to be handled differently.

This PR adds preliminary logic needed to distinguish between files originating from the auto or manual AutoTEM workflows, and will skip processing the file if it belongs to the manual AutoTEM workflow for now. Subsequent PRs will introduce logic to trigger processing and parse metadata correctly for the manual workflow.

### DIFF
--- a/src/murfey/client/contexts/fib.py
+++ b/src/murfey/client/contexts/fib.py
@@ -182,6 +182,23 @@ ACTIVITY_FIELD_MAP = (
 )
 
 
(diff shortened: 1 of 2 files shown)


stephen-riggs (approved):
Seems fine to me

Issue #278, opened 2024-05-21
LIF to TIFF conversion command line tool


d-j-hatton:
Needed for use with cluster submission

PR #837, opened 2026-07-03
Convert xrm images to tiffs

stephen-riggs:
This adds functionality to read xrm images and request these are converted to tiffs.

Currently this identifies atlas and roi images by finding out if they are montages, and splits them based on pixel size. Better methods may be possible but this should work for now.

Rather than doing the xrm to tiff conversion here it puts a message onto rabbitmq. See https://github.com/DiamondLightSource/cryoem-services/pull/278

Also contains a stub for registering rois in the database.

### DIFF
--- a/src/murfey/client/contexts/sxt.py
+++ b/src/murfey/client/contexts/sxt.py
@@ -178,10 +178,10 @@ class SXTContext(Context):
                 ):
                     x_tiles = _get_ole_header_value(
                         xrm_ole, "ImageInfo/XPosition", np.float32
-                    ).tolist()[0]
+                    ).tolist()
                     y_tiles = _get_ole_header_value(
                         xrm_ole, "ImageInfo/YPosition", np.float32
-                    ).tolist()[0]
+                    ).tolist()
                     metadata["x_position"] = x_tiles[int(len(x_tiles) / 2)]
                     metadata["y_position"] = y_tiles[int(len(y_tiles) / 2)]
 
--- a/src/murfey/client/contexts/sxt.py
+++ b/src/murfey/client/contexts/sxt.py
@@ -184,17 +184,17 @@ class SXTContext(Context):
                 if xrm_ole.exists("ImageInfo/PixelSize"):
                     metadata["pixel_size"] = np.frombuffer(
                         xrm_ole.openstream("ImageInfo/PixelSize").getvalue(), np.float32
-                    ).tolist()
+                    ).tolist()[0]
 
                 if xrm_ole.exists("ImageInfo/ImageHeight"):
                     metadata["height"] = np.frombuffer(
                         xrm_ole.openstream("ImageInfo/ImageHeight").getvalue(), np.int32
-                    ).tolist()
+                    ).tolist()[0]
 
                 if xrm_ole.exists("ImageInfo/ImageWidth"):
                     metadata["width"] = np.frombuffer(
-                        xrm_ole.openstream("ImageInfo/ImageWidth").getvalue(), np.intt32
-                    ).tolist()
+                        xrm_ole.openstream("ImageInfo/ImageWidth").getvalue(), np.int32
+                    ).tolist()[0]
 
                 # Find images which are not mosaics (txrm spec typos this as mosiac)
                 if xrm_ole.exists("ImageInfo/MosiacRows") and xrm_ole.exists(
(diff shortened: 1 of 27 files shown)


src/murfey/client/contexts/sxt.py
```diff
@@ -157,17 +157,176 @@ def post_transfer(
...
+                    Path(self._machine_config.get("rsync_basepath", "")),
+                )
+                if (
+                    environment.visit
+                    in Path(environment.default_destinations[source]).parts
+                ):
+                    # Split either side of the raw directory
+                    visit_idx = Path(
+                        environment.default_destinations[source]
+                    ).parts.index(environment.visit)
```
tieneupin, line 232: Maybe store `Path(environment.default_destinations[source]).parts` as a variable, since it's repeated quite a lot?

```
                if ... :
                    parts = Path(environment.default_destinations[source]).parts
                    visit_idx = parts.index(environment.visit)
                    destination_base = "/".join(parts[: visit_idx + 1]
                    destination_extra = "/".join(parts[visit_idx + 2 :]
                else:
                    ...
```

src/murfey/client/contexts/sxt.py
```diff
@@ -157,17 +157,176 @@ def post_transfer(
...
+                    },
+                )
+
+                if (
+                    metadata.get("mosaic_size", 1) > 0
+                    and metadata.get("pixel_size", 0) > 0.1
+                ):
+                    # Large pixel size, this is an atlas
+                    dcg_data = {
+                        "experiment_type_id": 44,  # Atlas
```
tieneupin, line 275: This will be fine when running here at our organisation, so not a critical issue, but it is technically dependent on the order in which the experiment types were inserted into ISPyB. The institution-independent way to do this would be to pass along the experiment type name, query the experiment type's experimentTypeId from the ISPyB in the `TransportManager` class, then insert that.

src/murfey/client/contexts/sxt.py
```diff
@@ -157,17 +157,176 @@ def post_transfer(
             **kwargs,
         )
 
-        data_suffixes = [".txrm"]
+        if transferred_file.suffix == ".xrm" and environment:
```
tieneupin, line 165: Checking `and environment` in each if-else clause is fine, but you could also consider having an early exit if no environment is found:
```
if not environment:
    logger.warning("No Murfey environment provided")
    return None
```

tieneupin (approved):
Some comments and suggestions, but the code looks good otherwise!

src/murfey/client/contexts/sxt.py
```diff
@@ -157,17 +157,176 @@ def post_transfer(
...
+                    y_tiles = np.frombuffer(
+                        xrm_ole.openstream("ImageInfo/YPosition").getvalue(), np.float32
+                    ).tolist()
+                    metadata["x_position"] = x_tiles[int(len(x_tiles) / 2)]
+                    metadata["y_position"] = y_tiles[int(len(y_tiles) / 2)]
+
+                if xrm_ole.exists("ImageInfo/PixelSize"):
+                    metadata["pixel_size"] = np.frombuffer(
+                        xrm_ole.openstream("ImageInfo/PixelSize").getvalue(), np.float32
+                    ).tolist()[0]
```
tieneupin: Personally, I think `np.frombuffer(xrm_ole.openstream("ImageInfo/PixelSize").getvalue(), np.float32)` could potentially be pulled out into a helper function, since it is repeated for every field:
```
def _get_value(ole_file: SomeType, path: str, dtype: NumpyDtype):
    return np.frombuffer(ole_file.openstream(path).getvalue(), dtype)
```

stephen-riggs: Done

src/murfey/client/contexts/sxt.py
stephen-riggs: I think this is fine as-is

src/murfey/client/contexts/sxt.py
stephen-riggs: True, but I think it's better not to do additional queries to ispyb. Using the type id is also less prone to typos either in ispyb or our code

src/murfey/client/contexts/sxt.py
stephen-riggs: I think this is fine as-is

PR #838, opened 2026-07-06
Initial model cube name fix

stephen-riggs:
I found I had this uncommitted from a while ago. It's a direct bug in the user-supplied initial model logic


tieneupin (approved):
Yep, looks fine! Does this urgently need to go in? Would it be too arduous to add a test for this?

PR #839, opened 2026-07-06
Look for references based on file naming

stephen-riggs:
There doesn't seem to be a nice metadata method for SXT references. This looks only for files with "ref" in the name


tieneupin (approved):
Nice, looks good to me!

PR #840, opened 2026-07-09
Add ispyb inserts for SXT atlases and ROIs

stephen-riggs:
Adds the functionality to insert SXT atlas and ROI images into both the murfey and ispyb databases. For ispyb the positions are adjusted to match the thumbnail images so that they display correctly on pato.

Requires atlas locations which weren't previously needed, so makes use of the `ImagingSite` table in addition to `DataCollectionGroup`. ROIs are inserted into the `SearchMap` table (matching GridSquare in ispyb).

Relatively low test coverage as I haven't written tests for the ispyb inserts.

### DIFF
--- a/src/murfey/client/contexts/sxt.py
+++ b/src/murfey/client/contexts/sxt.py
@@ -52,10 +52,14 @@ def _find_reference(txrm_file: Path) -> Path | None:
                     )[0]
                 )
         if mosaic_size == 0:
-            logger.info(f"Found reference {ref_option}")
+            logger.info(f"Found reference {ref_option.name}")
             return Path(ref_option.full_path)
     logger.warning(f"No reference found for {txrm_file}")
     return None
+
+
+def _get_ole_header_value(ole_file, title: str, dtype: np.dtype):
+    return np.frombuffer(ole_file.openstream(title).getvalue(), dtype)
 
 
 class SXTContext(Context):
@@ -125,9 +129,7 @@ class SXTContext(Context):
                 data=dc_data,
             )
 
-            recipes_to_assign_pjids = [
-                "sxt-aretomo",
-            ]
+            recipes_to_assign_pjids = self._machine_config.get("recipes", {}).values()
             for recipe in recipes_to_assign_pjids:
                 capture_post(
                     base_url=str(environment.url.geturl()),
@@ -158,6 +160,7 @@ class SXTContext(Context):
             environment=environment,
             **kwargs,
         )
+        metadata: dict[str, Any] = {}
 
         if transferred_file.suffix == ".xrm" and environment:
             # Make sure we have a dcg for this grid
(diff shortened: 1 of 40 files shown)


src/murfey/client/contexts/sxt.py
```diff
@@ -274,7 +277,7 @@ def post_transfer(
                     dcg_data = {
                         "experiment_type_id": 44,  # Atlas
                         "tag": dcg_tag,
-                        "atlas": str(converted_file_path),
+                        "atlas": str(thumbnail_path),
                         "atlas_pixel_size": round(metadata.get("pixel_size", 0), 2),
```
tieneupin, line 281: Would this need to be adjusted to account for the difference in pixel size between the original and the thumbnail image? I assume there will be some rescaling involved.

src/murfey/server/ispyb.py
```diff
@@ -543,6 +543,96 @@ def do_update_search_map(
...
+        if (
+            roi_parameters.pixel_size
+            and roi_parameters.height
+            and roi_parameters.height_on_atlas
+        ):
+            roi_parameters.pixel_size *= (
+                roi_parameters.height / roi_parameters.height_on_atlas
+            )
+        roi_parameters.x_location = (
+            int(roi_parameters.x_location8) if roi_parameters.x_location else None
```
tieneupin: Typo here.
```
int(roi_parameters.x_location8) --> int(roi_parameters.x_location)

File: src/murfey/workflows/sxt/sxt_metadata.py [hunk 3]
```diff
@@ -0,0 +1,127 @@
...
+        select(DataCollectionGroup)
+        .where(DataCollectionGroup.session_id == session_id)
+        .where(DataCollectionGroup.tag == roi_parameters.tag)
+    ).one()
+    roi_query = murfey_db.exec(
+        select(SearchMap)
+        .where(SearchMap.name == roi_name)
+        .where(SearchMap.tag == roi_parameters.tag)
+        .where(SearchMap.session_id == session_id)
+    ).all()
```
tieneupin: If only one result is to be expected under normal operating conditions for each combination of `name`, `tag`, and `session_id`, would it be better to use `.one_or_none()` and let it error if there's more than one matching entry?

src/murfey/workflows/sxt/sxt_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+            _transport_object.do_update_sxt_roi(roi.id, roi_parameters)
+    else:
+        logger.info(f"Registering new sxt roi {sanitise(roi_name)}")
+        if _transport_object:
+            roi_ispyb_response = _transport_object.do_insert_sxt_roi(
+                dcg.atlas_id, roi_parameters
+            )
+        else:
+            # mock up response so that below still works
+            roi_ispyb_response = {"success": False, "return_value": None}
```
tieneupin, line 49: I wonder if it will be shorter in this form:
```
roi_ispyb_response = _transport_object.do_insert_sxt_roi(
    dcg.atlas_id, roi_parameters
) if _transport_object else {"success": False, "return_value": None}
```

src/murfey/workflows/sxt/sxt_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        # Register new search map
+        roi = SearchMap(
+            id=(
+                roi_ispyb_response["return_value"]
+                if roi_ispyb_response["success"]
+                else None
+            ),
```
tieneupin, line 56: If the success code is just used to determine the return value, I believe the same effect can be achieved using `roi = SearchMap(id=roi_ispyb_response.get("return_value", None), ...)`

src/murfey/workflows/sxt/sxt_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+            y_stage_position=roi_parameters.y_stage_position,
+            pixel_size=roi_parameters.pixel_size,
+            width=roi_parameters.width,
+            height=roi_parameters.height,
+            image=roi_parameters.image or "",
+        )
+
+    atlas_sites = murfey_db.exec(
+        select(ImagingSite).where(ImagingSite.dcg_id == dcg.id)
+    ).all()
```
tieneupin: Same here. Will there be more than one atlas images registered for a given `dcg_id` in the SXT workflow? If no, you could switch to `.one_or_none()` and let it error otherwise.

src/murfey/server/api/workflow_sxt.py
```diff
@@ -62,23 +64,22 @@ def convert_xrm_to_tiff(xrm_file: XrmFile, db=murfey_db):
...
-    visit_name: str,
     session_id: MurfeySessionID,
-    tilt_series_info: SXTTiltSeriesInfo,
+    roi_name: str,
+    roi_info: SearchMapParameters,
     db=murfey_db,
 ):
-    # TODO
-    return
+    if _transport_object:
```
tieneupin, line 74: Worth logging the absence of `_transport_object`?

src/murfey/workflows/sxt/sxt_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+        ]
+    ):
+        atlas = atlas_sites[0]
+        # Convert from stage position to pixel locations
+        roi.x_location = (roi.x_stage_position - atlas.pos_x) / atlas.image_pixel_size
+        roi.y_location = (roi.y_stage_position - atlas.pos_y) / atlas.image_pixel_size
+
+        # Scaling from different pixel size of atlas and roi, and atlas thumbnail size
+        roi_parameters.x_location = roi.x_location * (512 / atlas.image_pixels_x) + 256
+        roi_parameters.y_location = 256 - roi.y_location * (512 / atlas.image_pixels_y)
```
tieneupin, line 89: Assuming array indices start from (0, 0) at the top-left of the image, does this mean that the y-axis goes from positive to negative as the array row number increases in the SXT workflow?
`roi_parameters.y_location = 256 - roi.y_location * (512 / atlas.image_pixels_y)` would mean that a negative `roi.y_location` corresponds to a `roi_parameters.y_location` value that's greater than 256.

src/murfey/workflows/sxt/sxt_metadata.py
```diff
@@ -0,0 +1,127 @@
...
+            )
+        )
+        roi_parameters.height_on_atlas = int(
+            round(
+                roi.height
+                * (roi.pixel_size / atlas.image_pixel_size)
+                * (512 / atlas.image_pixels_y)
+            )
+        )
+        if _transport_object:
```
tieneupin, line 104: Would it be worth logging the absence of a configured `_transport_object`?

tieneupin (changes requested):
One typo'ed attribute that needs fixing, and some optional suggestions.

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: Yes, I think that's reasonable. I was using `.all()` as it's a bit safer, but it's probably fine to let it fail if there is more than one

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: True, but the other similar instances in murfey are written in the form I've done

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: I think no to this. It's true that currently we only give a `return_value` when `success` is True, but that's a choice we make as they are independent. Therefore I think it's safest to check the `success` field

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: As above, agreed and done but maybe could cause issues

src/murfey/server/api/workflow_sxt.py
stephen-riggs: Done

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: Done

src/murfey/workflows/sxt/sxt_metadata.py
stephen-riggs: This is to match the conventions used in pato. When I was testing in the dev setup I concluded that it should be this way up to display properly. The `roi_parameters.y_location` should be between 0 and 512 for pato

src/murfey/client/contexts/sxt.py
stephen-riggs: Strictly yes, that's true the pixel size will not correspond to the image. As far as I'm aware the pixel size is never used when displaying the thumbnails though so I think this way is ok. It's not ideal to do any conversions here as the thumbnail size is set in cryoem-services and not in murfey

tieneupin (approved):
Thanks for addressing my comments! This should be good to merge now.

PR #841, opened 2026-07-14
Trigger cryoSIM processing

tieneupin:
Adds more functionality to the cryoSIM workflow.
* On the client side, Murfey will apply the cryoSIM context when it sees raw data files matching currently  known criteria (raw data files have no suffix, and end with a certain set of patterns).
* When a fluorescent data file is seen, it will submit a post to trigger data processing for the file.
* Adds an endpoint to the backend server that will receive requests to trigger data processing from the client, and will forward the message to RabbitMQ's `processing_recipe` queue. The actual submission of the message will be deferred until the processing recipe and service have been set up.
* Updates existing tests and adds new tests to reflect the updated behaviour.

Actual diff is about +170 lines, with the rest of the diff being due to tests.


src/murfey/client/contexts/sim.py
```diff
@@ -3,10 +3,39 @@
 
 from murfey.client.context import Context
 from murfey.client.instance_environment import MurfeyInstanceEnvironment
+from murfey.util.client import capture_post
 
 logger = logging.getLogger("murfey.client.contexts.sim")
 
 
+def _get_source(file_path: Path, environment: MurfeyInstanceEnvironment) -> Path | None:
```
stephen-riggs: You might be able to use the `_get_source` in the `context.py` as that should have the same behaviour. Looks like some contexts use that and some don't

src/murfey/client/contexts/sim.py
```diff
@@ -3,10 +3,39 @@
...
+    """
+    Returns the Path of the file on the client PC.
+    """
+    for s in environment.sources:
+        if file_path.is_relative_to(s):
+            return s
+    return None
+
+
+def _file_transferred_to(
```
stephen-riggs: Same here, `_file_transferred_to` exists in the parent `context.py` with the same behaviour

src/murfey/client/contexts/sim.py
```diff
@@ -3,10 +3,39 @@
...
+    environment: MurfeyInstanceEnvironment,
+    source: Path,
+    file_path: Path,
+    rsync_basepath: Path,
+) -> Path | None:
```
stephen-riggs: How would this return None?

src/murfey/client/contexts/sim.py
```diff
@@ -25,4 +54,56 @@ def post_transfer(
...
+            if source is None:
+                logger.warning(f"No source found for file {transferred_file}")
+                return None
+            destination_file = _file_transferred_to(
+                environment=environment,
+                source=source,
+                file_path=transferred_file,
+                rsync_basepath=Path(self._machine_config.get("rsync_basepath", "")),
+            )
+            if destination_file is None:
```
stephen-riggs: Same as above, what situation would the destination_file be None?

stephen-riggs (approved):
No issues I can see, but questions about code duplication and return values

src/murfey/client/contexts/sim.py
tieneupin: You're right. This would be a good opportunity to make sure that the shared `get_source` works across workflows. I'll switch to it and test it for the SIM workflow.

src/murfey/client/contexts/sim.py
tieneupin: This might have been oversight on my part back when I was learning how to set up the function for the CLEM workflow. I'll switch over to using the shared `file_transferred_to` function.

tieneupin:
Thanks for the quick review and helpful comments. I've taken your suggestions on board, and switched to using the shared `_get_source` and `_file_transferred_to` functions defined in `murfey.client.context`. The tests have passed, as has the test transfer on actual data in our dev setup, so I have gone ahead and merged this PR.

A future housekeeping-related PR would involve revisiting the other context files developed and switching them to using the shared `_get_source` and `_file_transferred_to` functions.

PR #842, opened 2026-07-16
Fix slot number determination logic for the FIB workflow

tieneupin:
In the Aquilos, the stage rotation value will be set to -75 degrees during normal milling and when acquiring SEM images. The xy coordinates at this rotation value are used to identify the slot number.

The xy coordinates needed to keep the same region in view will change as the stage is rotated, so a transformation matrix will need to be applied using the xy coordinates and the current rotation value to determine what the xy coordinates of the viewed region will be at the standard viewing/milling angle, and thus what slot number it belongs to.

This PR migrates the slot number determination logic out from the Pydantic model into separate helper functions. A `rotation_offset` dictionary key will be loaded from the `MachineConfig.calibrations` field, which can be used to transform the FIB stage positions into the correct frame of reference, from which the slot number can then be correctly identified. The tests have also been updated to reflect this.

### DIFF
--- a/src/murfey/util/models.py
+++ b/src/murfey/util/models.py
@@ -5,7 +5,7 @@ from datetime import datetime
 from pathlib import Path
 from typing import Any, Dict, List, Optional
 
-from pydantic import BaseModel, computed_field, field_validator
+from pydantic import BaseModel, field_validator
 
 """
 =======================================================================================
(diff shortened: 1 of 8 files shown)


src/murfey/util/models.py
```diff
@@ -111,9 +111,30 @@ class StagePositionValues(BaseModel):
...
+
+        The xy values that keep a particular point on the grid at the euncentric focus
+        under the electron and ion beams will change based on the stage rotation. This
+        means that a transformation matrix will need to be applied to estimate the
+        x-value of the region when the stage rotation is at -75 degrees. The correct
+        slot number can then be determined from there.
+        """
+        if self.x is not None and self.y is not None and self.rotation is not None:
+            # Rotate the xy-coordinates to the -75 degrees frame
+            theta = math.radians(self.rotation + 75)
```
stephen-riggs: I think ideally we want a way to be able to change the rotation value, with 75 as the default

stephen-riggs (approved):

src/murfey/util/models.py
tieneupin: I agree. I think we can try and discuss how that might be implemented as part of this PR before merging this.

tieneupin: After discussion, I've reworked the logic so that the slot number is determined using a helper function. This helper function can take in a `rotation_offset` parameter from the `MachineConfig`'s `calibrations` key, which would allow us to change the rotation value via updating the config.

d-j-hatton (approved):

PR #843, opened 2026-07-17
Skip processing for SXT stacks where all angles are zero

stephen-riggs:
It doesn't make sense to treat stacks as tomograms if there are no tilts!

TBC exactly what these are, but they appear to be focus scans on the SXT instrument. AreTomo is able to process them, but Imod is not.
Instead just do the txrm to tiff conversion for any stacks which contain only zero-degree tilts.

There is very little new code here, but a few bits have moved around to reuse the tiff path creation and get the right posts.

### DIFF
--- a/src/murfey/client/contexts/sxt.py
+++ b/src/murfey/client/contexts/sxt.py
@@ -76,6 +76,38 @@ class SXTContext(Context):
         self._basepath = basepath
         self._machine_config = machine_config
 
+    def determined_converted_tiff_path(
+        self,
+        transferred_file: Path,
+        environment: MurfeyInstanceEnvironment,
+        source: Path,
+    ):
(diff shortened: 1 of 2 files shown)


src/murfey/client/contexts/sxt.py
```diff
@@ -76,6 +76,38 @@ def __init__(
         self._basepath = basepath
         self._machine_config = machine_config
 
+    def determined_converted_tiff_path(
```
tieneupin: Does this need to be in the past tense? Functions are normally present tense, no?

tieneupin (approved):
The new class function's name is in the past tense, but beyond that, everything looks good!

src/murfey/client/contexts/sxt.py
stephen-riggs: This was a typo, now fixed

PR #846, opened 2026-07-28
Request processing of lamella evaluation images

tieneupin:
Lamella evaluation images are high-resolution images of the sites during the last stages of the milling process. These images should be the ones displayed on the PATo frontend when users click on regions of interest on the atlas, in a similar vein to the CLEM workflow.

This PR adds logic to the `FIBContext` to submit a POST request to the backend when it detects a lamella evaluation image. An API endpoint and a workflow module have also been set up on the backend to eventually process the received message. The processing logic will be refined in subsequent PRs, and currently just exists as a placeholder.

Actual code diff is +117/-21, with the rest of the changes being due to the addition and modification of tests.


d-j-hatton (approved):
looks fine

PR #848, opened 2026-07-30
Add lamella tomography experiment type id

stephen-riggs:
ispyb now has a different experiment type id for lamella tomography compared to grid tomo

### DIFF
--- a/src/murfey/client/context.py
+++ b/src/murfey/client/context.py
@@ -124,10 +124,10 @@ def ensure_dcg_exists(
 
         if collection_type == "tomo":
             windows_path = session_data["TomographySession"]["AtlasId"]
-            lamella_workflow = session_data.get("TomographySession", {}).get(
-                "LamellaWorkflow", None
-            )
-            if lamella_workflow == "true":
+            if (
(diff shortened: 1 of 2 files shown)


src/murfey/client/context.py
```diff
@@ -124,6 +124,14 @@ def ensure_dcg_exists(
 
         if collection_type == "tomo":
             windows_path = session_data["TomographySession"]["AtlasId"]
+            lamella_workflow = session_data.get("TomographySession", {}).get(
+                "LamellaWorkflow", None
+            )
+            if lamella_workflow == "true":
+                # Lamella tomography experiment type
+                experiment_type_id = 49
```
tieneupin: `lamella_workflow` seems to only be used once. Could perhaps be shortened to:
```
if session_data.get("TomographySession", {}).get("LamellaWorkflow", None) == "true":
    experiment_type_id = 49
else:
    ...
```

tieneupin (approved):
Optional suggestion for trimming down the number of lines, but other than that, looks good. Any chance of adding tests for the lamella tomography workflow as part of this PR?

PR #850, opened 2026-07-31
Workarounds in reposting failed calls

stephen-riggs:
For some reason the `repost_failed_calls` function does not always pick up the type correctly. This addresses that by doing a lookup if the type ends up as a string.

Also for grid squares takes the last one, to partly address the problems when grid slots are duplicated.


tieneupin (approved):
Looks good to me!

PR #851, opened 2026-08-04
Allow reprocessing of files for SXT

stephen-riggs:
Do not skip processing for SXT files which were already processed, as they may have a new reference attached.

This will trigger reprocessing on session reconnection, which isn't ideal but is probably ok for now

### DIFF
--- a/src/murfey/workflows/sxt/process_sxt_tilt_series.py
+++ b/src/murfey/workflows/sxt/process_sxt_tilt_series.py
@@ -48,7 +48,6 @@ def process_sxt_tilt_series(
         tilt_series = tilt_series_query[0]
         if tilt_series.processing_requested:
             logger.info(f"Tilt series {tilt_series.tag} has already been processed")
-            # return {"success": True}
     else:
         tilt_series = TiltSeries(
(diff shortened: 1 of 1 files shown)


src/murfey/workflows/sxt/process_sxt_tilt_series.py
```diff
@@ -48,7 +48,7 @@ def process_sxt_tilt_series(
         tilt_series = tilt_series_query[0]
         if tilt_series.processing_requested:
             logger.info(f"Tilt series {tilt_series.tag} has already been processed")
-            return {"success": True}
+            # return {"success": True}
```
d-j-hatton: you can probably just remove the line rather than comment it out to save a little bit of fluff build-up. undoing the change will just involve rolling back this pr

d-j-hatton (approved):

src/murfey/workflows/sxt/process_sxt_tilt_series.py
stephen-riggs: yes, that seems sensible

Issue #291, opened 2024-05-24
Some buttons on the Murfey GUI don't work on Leica PC


tieneupin:
Buttons that don't work:
* Visit complete
* Remove session

PR #855, opened 2026-08-05
Update SIM processing workflow

tieneupin:
Having fleshed out the PySIMRecon service in `cryoem-services` (see https://github.com/DiamondLightSource/cryoem-services/pull/286 and https://github.com/DiamondLightSource/cryoem-services/pull/291), this PR updates the `request_sim_processing` endpoint so that it includes the correct parameters in the outbound message.

* Renamed `request_sim_processing` API endpoint to `request_sim_reconstruction`
* Updated the `SIMContext` logic so that only SIM files ending with `"_FL"` are processed; raw data files without `"_FL"` will be transferred but skipped
* Updated parameters in the outbound message
* Updated tests

### DIFF
--- a/src/murfey/server/api/workflow_sim.py
+++ b/src/murfey/server/api/workflow_sim.py
@@ -27,8 +27,8 @@ class SIMDataFile(BaseModel):
     file: Path
 
 
-@router.post("/sessions/{session_id}/process_data")
-def request_sim_processing(
+@router.post("/sessions/{session_id}/sim_recon")
+def request_sim_reconstruction(
     session_id: int,
     sim_data: SIMDataFile,
     murfey_db: SQLModelSession = murfey_db,
--- a/src/murfey/server/api/workflow_sim.py
+++ b/src/murfey/server/api/workflow_sim.py
@@ -1,13 +1,18 @@
 import json
 import logging
 from pathlib import Path
+from typing import Any
 
 from fastapi import APIRouter, Depends
 from pydantic import BaseModel
+from sqlmodel import Session as SQLModelSession, select
 
+import murfey.util.db as MurfeyDB
 from murfey.server import _transport_object
 from murfey.server.api.auth import validate_instrument_token
+from murfey.server.murfey_db import murfey_db
 from murfey.util import sanitise_path
+from murfey.util.config import get_machine_config
 
 logger = logging.getLogger("murfey.server.api.workflow_sim")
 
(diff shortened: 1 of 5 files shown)


src/murfey/server/api/workflow_sim.py
```diff
@@ -22,22 +27,69 @@ class SIMDataFile(BaseModel):
...
+        return None
+
+    # Load PySIMRecon values from the machine config
+    try:
+        machine_config = get_machine_config(instrument_name)[instrument_name]
+        pysimrecon_config: dict[str, dict[str, Any]] | None = (
+            machine_config.calibrations.get("pysimrecon_config")
+        )
+        if not pysimrecon_config:
+            logger.error(f"No PySIMRecon configuration found for {instrument_name}")
```
stephen-riggs: If there aren't parameters, do you need to reject the processing or can you just let the service use defaults?

src/murfey/server/api/workflow_sim.py
```diff
@@ -22,22 +27,69 @@ class SIMDataFile(BaseModel):
...
+        return None
+
     # Construct message and submit it to 'processing_recipe'
     logger.info(
         f"Submitting request to process the cryoSIM file {sanitise_path(sim_data.file)}"
     )
+    # Construct the output directory for the PySIMRecon outputs to be saved to
+    try:
+        visit_idx = sim_data.file.parent.parts.index(visit_name)
+        raw_dir = sim_data.file.parents[-(visit_idx + 2)]
```
stephen-riggs: The `-(visit_idx + 2)` is correct, but confused me a lot initially

src/murfey/server/api/workflow_sim.py
```diff
@@ -22,22 +27,69 @@ class SIMDataFile(BaseModel):
...
+        output_dir = (
```
stephen-riggs, line 102: There's probably as easier way to do this - I think it's just a substitution of `processed` for `raw*`?

src/murfey/server/api/workflow_sim.py
tieneupin: Yes, I can appreciate that. `Path().parents` arranges the parent paths in descending order (immediate parent to root), so finding the parent based on index will have to start from the end of the list.

src/murfey/server/api/workflow_sim.py
tieneupin: Yes, that would be the most direct way. I have probably overengineered it slightly, but this method would allow for the code to still work if we decide to use a directory name other than `"raw"` in the future.

src/murfey/server/api/workflow_sim.py
tieneupin: That makes sense. I could provide the latest wavelength values we currently have as the default values, and the ones read from the MachineConfig could be updated as and when the instrument's calibrations are changed.

